PR #2, opened 2024-02-19
Migrate Anbox Streaming SDK to GitHub

adglkh:


js/LICENSE
morphis, line 1: Why do we have this a second time? We should have this once at the top of the repository?

examples/scripts/build-with-docker.sh
```diff
@@ -0,0 +1,145 @@
+#!/bin/sh -ex
+#
+# Copyright 2020 Canonical Ltd.  All rights reserved.
+#
```
morphis, line 4: We need to update the copyright header to reflect Apache-2.0

examples/scripts/clean-build.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/sh -ex
+#
+# Copyright 2020 Canonical Ltd.  All rights reserved.
+#
```
morphis, line 4: See above

js/run-tests.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh -e
+#
+# Anbox Streaming SDK
+# Copyright 2021 Canonical Ltd.  All rights reserved.
```
morphis, line 4: See above

js/tests/unit/run-tests.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/sh -ex
+#
+# Anbox Streaming SDK
+# Copyright 2021 Canonical Ltd.  All rights reserved.
+#
```
morphis, line 5: See above

scripts/build.sh
```diff
@@ -0,0 +1,108 @@
+#!/bin/sh -xe
+#
+# Anbox Streaming SDK
+# Copyright 2019 Canonical Ltd.  All rights reserved.
```
morphis, line 4: See above

scripts/gen-version.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh -e
+#
+# Anbox Streaming Stack
+# Copyright 2020 Canonical Ltd.  All rights reserved.
```
morphis, line 4: See above

scripts/run-tests.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash -ex
+#
+# Copyright 2020 Canonical Ltd.  All rights reserved.
```
morphis, line 3: See above

scripts/streaming-sdk-files.allowlist
morphis, line 1: This is no longer needed

scripts/validate.sh
morphis, line 1: This is no longer needed

morphis (changes requested):

js/LICENSE
adglkh: I think the original idea is to create a release tarball without cp the license around(from the top to the nested folder)
I can drop this for simplicity.

scripts/streaming-sdk-files.allowlist
adglkh: To me, the `streaming-sdk-files.allowlist` and the following `scripts/validate.sh` are still required to ensure there are no other pieces of needless files to be injected into the resulting zip tarball.
The content of the resulting zip tarball won't include every piece of source files in this repo(E.g. scripts/*, example/scripts/*) after this PR is merged and it will remain the same as the current layout.

examples/scripts/build-with-docker.sh
adglkh: Thanks for pointing this out!
Will fix.

examples/js/anbox-stream-sdk.js
jat-canonical, line 1: Has this been moved under `js/`? as I cannot see it there.

examples/scripts/build-with-docker.sh
jat-canonical, line 1: Examples of using the sdk make sense but why do we need example scripts for building it ?

jat-canonical (commented):
A few comments

examples/js/anbox-stream-sdk.js
adglkh: The layout in the resulting release tarball/artifact is identical to what we have today(E..g [the artifact]( https://github.com/adglkh/anbox-streaming-sdk/actions/runs/7999651120) includes all contents after this change) but not 1:1 mapping from the source layout in the repo after this MP got merged.

I'd suggest taking a look AC074 spec for details.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,44 @@
...
+    branches:
+      - '**'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
```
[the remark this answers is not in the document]
adglkh: Okay, I can change that later on.

examples/js/anbox-stream-sdk.js
jat-canonical: yes but what i mean is i cannot see the js source files at all under js.

adglkh: We only ship one piece of js file in the trunk, and cp it to each js folder when creating a release tarball.
In this way, if any code change is made to the js file, we won't replicate the same change to other copy of js files.
See:
https://github.com/canonical/anbox-streaming-sdk/pull/2/files#diff-3d64996115d0ac3b74e36f57ce6b00d6b4982614c41b1ed39eab8ef43b2e9a2bR68

jat-canonical: alright makes sense

scripts/build.sh
```diff
@@ -0,0 +1,122 @@
...
+for p in "$@"
+do
+  case $p in
+    --proxy=*)
+        PROXY=${p#*=}
+        ;;
+    --version=*)
+        VERSION=${p#*=}
+        ;;
+    --create-tarball)
```
morphis, line 30: As mentioned above I don't we need to create a tarball at all. What do we win by doing that? If correctly tag the git repository will automatically create an archive for each release for us without us doing anything,

morphis (changes requested):

examples/js/anbox-stream-sdk.js
adglkh: 1. Using the git archive feature provided by Github results in creating a zip tarball for the entire repository, which includes quite a few needless contents, such as scripts/. After this PR is merged, the content layout of this repository varies and is not identical to what we want to deliver in our release tarball (Check what we currently have in this repo vs the sources in the PR). 
2. With the tarball archived with git, which missing js SDK file in all examples. users have to manually copy the js sdk file into each example repository before building and running it, which is not straight forward for the end user (given that we only maintain one piece of js file in this repository.)
3. For the same reason in 2), users must also build the AAR and APK files themselves before being able to play with Android examples.

Hence as a release tarball, the out-of-the-box experience is not satisfactory
That's why the `--create-tarball` cli parameter is introduced into the build script to generate the proper release tarball, align with what we deliver today.

examples/scripts/build-with-docker.sh
adglkh: We can do that.

morphis (approved):
LGTM lets merge to get a first version in place. Any cleanup can happen afterwards.

PR #18, opened 2024-03-05
feat(stream-sdk): remove API fallback mechanism

ajanon:
Version 1 of the API was used by Anbox until 1.17.2, 10 months ago. That's out of standard support for a while.

This patch comes directly from our internal repository and was apparently missed when moving to GH.


adglkh (approved):
LGTM

morphis (approved):
LGTM

PR #21, opened 2024-03-20
feat(js): add support for pointer lock in the SDK

morphis:
With support for pointer lock we can support use cases where the mouse and relative motion information is used instead of absoluate coordinates, like in various games not using touch input but explicit mouse information. Support is kept behind a feature flag until we have stablized and tested the implementation further.


adglkh (approved):
LGTM

jat-canonical (approved):
LGTM

ajanon (approved):
LGTM

PR #22, opened 2024-04-08
feat(stream-sdk): add vhalReady callback to notify of VHAL availability

lorumic:
Added `vhalReady` callback to allow a consumer of the streaming SDK to be notified of the discovery of VHAL capabilities for the stream.


adglkh (approved):
LGTM

ajanon (approved):
LGT

PR #25, opened 2024-04-18
fix(js): sync up the JS sdk file

adglkh:
In the following commit(4b36ac770091abbfe362013310c3c7d8f2ab0c95)
```
feat(stream-sdk): Remove API fallback mechanism

Version 1 of the API was used by Anbox until 1.17.2, 10 months ago. That's out
of standard support for a while.
```

We removed the API fallback mechanism in the JS file. However, the
code change was only made to the following JS sdk file in the following
folder:

examples/android/enhanced_webview_streaming/app/src/main/assets/js/

This directory is not where the JS SDK file is supposed to be located.
We only maintain one piece of js SDK file, which should always be located
in the `js` folder, and copy it around when building the release tarball.

This change fixed that.

### DIFF
--- a/js/anbox-stream-sdk.js
+++ b/js/anbox-stream-sdk.js
@@ -18,9 +18,6 @@
  * limitations under the License.
  */
 
-// Error code returned when an unknown signaling request has been received
-const ANBOX_STREAM_SIGNALING_ERROR_BAD_REQUEST = 5;
-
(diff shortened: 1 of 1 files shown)


jat-canonical:
LGTM

jat-canonical (approved):

ajanon (approved):
LGTM, thanks!

PR #28, opened 2024-04-25
feat(video): allow to change default vertically centered video alignment

lorumic:
Added an optional flag to the `AnboxStream` options object to opt out of the default vertically centered alignment for the video element inside its container. For backward compatibility, it will be kept vertically centered by default, so that it doesn't interfere with previous usages by causing unexpected changes.

### DIFF
--- a/js/anbox-stream-sdk.js
+++ b/js/anbox-stream-sdk.js
@@ -75,6 +75,7 @@ class AnboxStream {
    * @param options: {object}
    * @param options.connector {object} WebRTC Stream connector.
    * @param options.targetElement {string} ID of the DOM element to attach the video to.
+   * @param [options.verticalAlignment=center] {top | center | bottom} Vertical alignment of the video element in its container.
    * @param [options.fullScreen=false] {boolean} Stream video in full screen mode.
    * @param [options.deviceType] {string} Send the type of device the SDK is running on to the Android container.
    * @param [options.enableStats] {boolean} Enable collection of statistics. Not recommended in production.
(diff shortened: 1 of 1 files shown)


ajanon (approved):
LGTM

ajanon:
One question, as we have now branched for 1.22 here: does this need to be backported? Should this be for 1.22.1 or 1.23?

js/anbox-stream-sdk.js
```diff
@@ -1194,9 +1198,9 @@ class AnboxStream {
       case 180:
         visualElement.style.height = playerHeight.toString() + "px";
         visualElement.style.width = playerWidth.toString() + "px";
-        visualElement.style.top = `${Math.round(
-          container.clientHeight / 2 - playerHeight / 2
-        )}px`;
+        visualElement.style.top = this._options.centerVertically
```
adglkh, line 1201: With the newly introduced boolean type`options.centerVertically`, it affects how the video element is positioned: either vertically centered inside the video container or aligned to the top of the container. However, if another requirement comes later, such as aligning the video element to the bottom of the container, the option may not cover this case well. Given that I would suggest using an option like `verticalAlignment` instead. This option could accept parameters such as `top`, `center`(default), and `bottom`, which would accommodate almost all cases we aim to support.

What do you think?

lorumic: Sure, we could do that. Should we also have one also for horizontal alignment at this point? To cover all the customisation possibilities.

ajanon:
> > One question, as we have now branched for 1.22 here: does this need to be backported? Should this be for 1.22.1 or 1.23?
> 
> I'd say 1.22.1. This is a mostly harmless change, as it is fully backward-compatible and not changing the previous default behaviour. Targeting 1.22.1 gives us more flexibility about when to release the visually-polished version of VHAL, since this is a prerequisite to address one of the changes requested there.

Then, if you don't mind, I will keep this open for now and merge it after release only.

lorumic:
> Then, if you don't mind, I will keep this open for now and merge it after release only.

Sure, no problem. We can also continue improving it if further ideas come. :)

js/anbox-stream-sdk.js
adglkh: We can add that at a later point if necessary. For now, keeping it as it is works fine for me.

adglkh (approved):
LGTM

PR #30, opened 2024-05-28
feat(e2e-tests): added e2e tests (setup, teardown and join session spec) [WD-11275]

lorumic:
Added e2e tests: setup, teardown and join session spec.

Fixes [WD-11275](https://warthogs.atlassian.net/browse/WD-11275)

[WD-11275]: https://warthogs.atlassian.net/browse/WD-11275?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -82,6 +82,7 @@ jobs:
       shell: bash
       run: |
         set -x
+        echo "CI=true" >> .env.local
         echo "ARCHITECTURE=amd64" >> .env.local
         echo "AMS_API_CERTIFICATE=anbox-cloud.crt" >> .env.local
         echo "AMS_API_CERTIFICATE_KEY=anbox-cloud.key" >> .env.local
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -82,6 +82,7 @@ jobs:
       shell: bash
       run: |
         set -x
+        echo "ARCHITECTURE=amd64" >> .env.local
         echo "AMS_API_CERTIFICATE=anbox-cloud.crt" >> .env.local
         echo "AMS_API_CERTIFICATE_KEY=anbox-cloud.key" >> .env.local
         echo "AMS_API_URL=$(sudo cat /var/snap/microbox/common/dashboard/config.yaml | grep AMS_API_URL | cut -d ' ' -f2)" >> .env.local
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -22,7 +22,7 @@ defaults:
 
 jobs:
   e2e:
-    name: e2e-tests
+    name: microbox, amd64
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@main
@@ -36,6 +36,20 @@ jobs:
         set -x
         openssl req -x509 -newkey rsa:4096 -keyout anbox-cloud.key -out anbox-cloud.crt -days 365 -nodes -subj "/C=GB/ST=London/L=London/O=Global Security/OU=IT Department/CN=example.com"
 
+    - name: Install LXD
+      uses: canonical/setup-lxd@v0.1.1
+      with:
+        channel: 5.0/stable
+
+    - name: Configure LXD
+      shell: bash
+      run: |
+        set -x
+        sudo lxc config set core.https_address "[::]:8443"
+        sudo lxc config set cluster.https_address "127.0.0.1:8443"
+        sudo lxc cluster enable local
+        sudo lxc storage create local zfs
+
     - name: Install microbox
       run: |
         set -x
@@ -47,25 +61,9 @@ jobs:
         sudo snap install --edge microbox
         sudo microbox init --auto
 
-    - name: Install LXD
-      uses: canonical/setup-lxd@v0.1.1
-      with:
-        channel: 5.0/stable
-
-    - name: Configure LXD
-      shell: bash
-      run: |
-        set -x
-        sudo lxc config set core.https_address "[::]:8443"
-        sudo lxc config trust add --name=microbox /var/snap/microbox/common/ams/lxd/client.crt
-        sudo lxc config set cluster.https_address "127.0.0.1"
-        sudo lxc cluster enable local
-        sudo lxc storage create local zfs
-
     - name: Configure AMC
       run: |
         set -x
-        sudo microbox.amc node add --unmanaged --storage-pool=local local 127.0.0.1
         sudo cp anbox-cloud.crt /root/client.crt
         sudo microbox.amc config trust add /root/client.crt
         sudo microbox.amc config set container.security_updates false
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -1,0 +1,103 @@
+name: E2e tests
+
+on:
+  workflow_call:
+    secrets:
+      UBUNTU_PRO_TOKEN:
+        required: true
+  workflow_dispatch:
+  schedule:
+    # execute the action nightly
+    - cron: '0 10 * * *'
+  push:
+    branches:
+      - '**'
+  pull_request:
+    branches:
+      - '**'
+
+defaults:
+  run:
+    working-directory: js/tests/e2e
+
+jobs:
+  e2e:
+    name: e2e-tests
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@main
+
+    - name: Store current date and time in output
+      id: datetime
+      run: echo "datetime=$(date -u +'%Y-%m-%dT%H.%M.%SZ')" >> $GITHUB_OUTPUT
+
+    - name: Generate keys
+      run: |
+        set -x
+        openssl req -x509 -newkey rsa:4096 -keyout anbox-cloud.key -out anbox-cloud.crt -days 365 -nodes -subj "/C=GB/ST=London/L=London/O=Global Security/OU=IT Department/CN=example.com"
+
+    - name: Install microbox
+      run: |
+        set -x
+        sudo pro attach ${{ secrets.UBUNTU_PRO_TOKEN }}
+        sudo pro enable --access-only anbox-cloud
+        sudo apt update
+        sudo apt install -y anbox-modules-dkms-122 linux-modules-extra-$(uname -r)
+        for n in binder_linux anbox_sync virt_wifi ; do sudo modprobe "$n" ; done
+        sudo snap install --edge microbox
+        sudo microbox init --auto
+
+    - name: Install LXD
+      uses: canonical/setup-lxd@v0.1.1
+      with:
+        channel: 5.0/stable
+
+    - name: Configure LXD
+      shell: bash
+      run: |
+        set -x
+        sudo lxc config set core.https_address "[::]:8443"
+        sudo lxc config trust add --name=microbox /var/snap/microbox/common/ams/lxd/client.crt
+        sudo lxc config set cluster.https_address "127.0.0.1"
+        sudo lxc cluster enable local
+        sudo lxc storage create local zfs
+
+    - name: Configure AMC
+      run: |
+        set -x
+        sudo microbox.amc node add --unmanaged --storage-pool=local local 127.0.0.1
+        sudo cp anbox-cloud.crt /root/client.crt
+        sudo microbox.amc config trust add /root/client.crt
+        sudo microbox.amc config set container.security_updates false
+        sudo microbox.amc config set registry.mode manual
+        sudo microbox.amc config set images.version_lockstep false
+        sudo microbox.amc config set images.url https://images.anbox-cloud.io/stable/
+        # NOTE We need to hide the token here to not accidentially leak credentials
+        set +x
+        sudo microbox.amc config set images.auth "bearer:$(sudo cat /var/lib/ubuntu-advantage/private/machine-token.json | jq -r '.resourceTokens[] | select(.type=="anbox-images").token')"
+        set -x
+        sudo microbox.amc image add jammy:android13:amd64 jammy:android13:amd64
+
+    - name: Configure env variables
+      shell: bash
+      run: |
+        set -x
+        echo "AMS_API_CERTIFICATE=anbox-cloud.crt" >> .env.local
+        echo "AMS_API_CERTIFICATE_KEY=anbox-cloud.key" >> .env.local
+        echo "AMS_API_URL=$(sudo cat /var/snap/microbox/common/dashboard/config.yaml | grep AMS_API_URL | cut -d ' ' -f2)" >> .env.local
+        echo "ASG_API_URL=$(sudo cat /var/snap/microbox/common/dashboard/config.yaml | grep ASG_API_URL | cut -d ' ' -f2)" >> .env.local
+        # NOTE We need to hide the token here to not accidentially leak credentials
+        set +x
+        echo "ASG_API_TOKEN=$(sudo cat /var/snap/microbox/common/dashboard/config.yaml | grep ASG_API_TOKEN | cut -d ' ' -f2)" >> .env.local
+        set -x
+
+    - name: Run tests
+      run: |
+        ./run-tests.sh
+
+    - uses: actions/upload-artifact@v4
+      if: always()
+      with:
+        name: playwright-report-${{ steps.datetime.outputs.datetime }}
+        path: js/tests/e2e/playwright-report/
+        retention-days: 30
--- a/js/tests/e2e/eslint.config.mjs
+++ b/js/tests/e2e/eslint.config.mjs
@@ -1,0 +1,28 @@
+/*
+ * This file is part of Anbox Cloud Streaming SDK
+ *
+ * Copyright 2024 Canonical Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import globals from "globals";
+import pluginJs from "@eslint/js";
+import eslintConfigPrettier from "eslint-config-prettier";
+
+export default [
+  { languageOptions: { globals: { ...globals.browser, ...globals.node } } },
+  { ignores: [".cache/"] },
+  pluginJs.configs.recommended,
+  eslintConfigPrettier,
+];
--- a/js/tests/e2e/global-setup.js
+++ b/js/tests/e2e/global-setup.js
@@ -18,6 +18,7 @@
 
 import { chromium, expect } from "@playwright/test";
 import { APP_NAME, BASE_URL } from "./fixtures/constants.cjs";
+require("dotenv").config({ path: ".env.local" });
 
 const isTestAppAvailable = async () => {
   const response = await fetch(`${BASE_URL}/asgApplications`);
--- a/js/tests/e2e/global-setup.js
+++ b/js/tests/e2e/global-setup.js
@@ -1,0 +1,135 @@
+/*
+ * This file is part of Anbox Cloud Streaming SDK
+ *
+ * Copyright 2024 Canonical Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { chromium, expect } from "@playwright/test";
+import { APP_NAME, BASE_URL } from "./fixtures/constants.cjs";
+
+const isTestAppAvailable = async () => {
+  const response = await fetch(`${BASE_URL}/asgApplications`);
+  const applications = await response.json();
+  return applications.find((application) => application.name === APP_NAME);
+};
+
+const hasTestSession = async () => {
+  const response = await fetch(`${BASE_URL}/sessions`);
+  const sessions = await response.json();
+  return sessions.find((session) => session.app === APP_NAME);
+};
+
+const hasActiveTestSession = async () => {
+  const response = await fetch(`${BASE_URL}/sessions`);
+  const sessions = await response.json();
+  const session = sessions.find(
+    (session) => session.app === APP_NAME && session.status === "active",
+  );
+  return session ? session.id : false;
+};
+
+const waitForSessionActive = async (page) => {
+  let retryCount = 0;
+  while (retryCount < 20) {
+    const sessionId = await hasActiveTestSession();
+    if (sessionId) return sessionId;
+    await page.waitForTimeout(3_000);
+    retryCount++;
+  }
+  return false;
+};
+
+const createTestSession = async () => {
+  const createSessionResponse = await fetch(
+    `${BASE_URL}/session?app=${APP_NAME}`,
+    {
+      method: "POST",
+    },
+  );
+  expect(createSessionResponse.status).toBe(200);
+};
+
+const hasTestApplication = async () => {
+  const response = await fetch(`${BASE_URL}/applications`);
+  const applications = await response.json();
+  return applications.some((application) => application.name === APP_NAME);
+};
+
+const hasReadyTestApplication = async () => {
+  const response = await fetch(`${BASE_URL}/applications`);
+  const applications = await response.json();
+  const app = applications.find((application) => application.name === APP_NAME);
+  if (app && app.status === "error") {
+    throw new Error("Application creation failed");
+  }
+  return applications.some(
+    (application) =>
+      application.name === APP_NAME && application.status === "ready",
+  );
+};
+
+const waitForAppReady = async (page) => {
+  let retryCount = 0;
+  // 1) check that status becomes ready
+  // we allow many retries as we need to give it time to download the image
+  while (retryCount < 200) {
+    const isReady = await hasReadyTestApplication();
+    if (isReady) break;
+    await page.waitForTimeout(3_000);
+    retryCount++;
+  }
+  // 2) check that application is available on ASG
+  while (retryCount < 200) {
+    const isAvailable = await isTestAppAvailable();
+    if (isAvailable) return true;
+    await page.waitForTimeout(3_000);
+    retryCount++;
+  }
+  return false;
+};
+
+const createTestApplication = async () => {
+  const createAppResponse = await fetch(`${BASE_URL}/application`, {
+    method: "POST",
+  });
+  expect(createAppResponse.status).toBe(200);
+};
+
+async function globalSetup() {
+  const hasApp = await hasTestApplication();
+  if (!hasApp) {
+    await createTestApplication();
+  }
+  const browser = await chromium.launch();
+  const page = await browser.newPage();
+  const isAppReady = await waitForAppReady(page);
+  if (!isAppReady) {
+    throw new Error("Application did not become ready");
+  }
+  const hasSession = await hasTestSession();
+  if (!hasSession) {
+    await createTestSession();
+  }
+  const sessionId = await waitForSessionActive(page);
+  if (!sessionId) {
+    throw new Error("Session did not become active");
+  }
+  process.env.SESSION_ID = sessionId;
+  // wait a few more seconds to let Android boot up
+  await page.waitForTimeout(10_000);
+  await browser.close();
+}
+
+export default globalSetup;
--- a/js/tests/e2e/global-teardown.js
+++ b/js/tests/e2e/global-teardown.js
@@ -1,0 +1,41 @@
+/*
+ * This file is part of Anbox Cloud Streaming SDK
+ *
+ * Copyright 2024 Canonical Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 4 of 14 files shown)


.github/workflows/ci.yaml
```diff
@@ -24,6 +28,74 @@ jobs:
         # See `scripts/gen-version.sh`
         fetch-depth: "0"
 
+    - name: Generate keys and allow permissions (for e2e tests)
+      working-directory: js/tests/e2e
+      run: |
+        set -x
+        openssl req -x509 -newkey rsa:4096 -keyout anbox-cloud.key -out anbox-cloud.crt -days 365 -nodes -subj "/C=GB/ST=London/L=London/O=Global Security/OU=IT Department/CN=example.com"
+        sudo chmod 0666 anbox-cloud.crt
+        sudo chmod 0666 anbox-cloud.key
```
morphis, line 37: I am not sure why you need 0666 here. Aren't the tests running with the same user you create the key/cert here?

.github/workflows/ci.yaml
```diff
@@ -24,6 +28,74 @@ jobs:
...
+
+    - name: Install microbox (for e2e tests)
+      run: |
+        set -x
+        sudo pro attach ${{ secrets.UBUNTU_PRO_TOKEN }}
+        sudo pro enable --access-only anbox-cloud
+        sudo apt update
+        sudo apt install -y anbox-modules-dkms-121 linux-generic-hwe-22.04
```
morphis, line 45:
```suggestion
        sudo apt install -y anbox-modules-dkms-122 linux-generic-hwe-22.04
```

It would be even better if we can calculate the package name based on `.base_version`

.github/workflows/ci.yaml
```diff
@@ -24,6 +28,74 @@ jobs:
...
+        sudo snap install --edge microbox
+        sudo snap set microbox ams.test-mode=true
+        yes '' | sudo microbox init
+
+    - name: Install LXD (for e2e tests)
+      uses: canonical/setup-lxd@v0.1.1
+      with:
+        channel: latest/edge
```
morphis, line 53: Why latest/edge? We currently only support 5.0 LTS. One of the next releases will add 5.21 support.

.github/workflows/ci.yaml
```diff
@@ -24,6 +28,74 @@ jobs:
...
+
+    - name: Configure LXD (for e2e tests)
+      shell: bash
+      run: |
+        set -x
+        sudo lxc config set core.https_address "[::]:8443"
+        sudo lxc config trust add --name=microbox /var/snap/microbox/common/ams/lxd/client.crt
+        sudo lxc config set cluster.https_address "127.0.0.1"
+        sudo lxc cluster enable local
+        sudo lxc storage create local dir
```
morphis, line 63: Why using `dir` storage backend here? Is that faster on the Github/self hosted runners?

.github/workflows/ci.yaml
```diff
@@ -24,6 +28,74 @@ jobs:
...
+        sudo cp anbox-cloud.crt /root/client.crt
+        sudo microbox.amc config trust add /root/client.crt
+        sudo microbox.amc config set registry.mode manual
+        sudo microbox.amc config set images.version_lockstep false
+        sudo microbox.amc config set images.url https://images.anbox-cloud.io/stable/
+        # NOTE We need to hide the token here to not accidentially leak credentials
+        set +x
+        sudo microbox.amc config set images.auth "bearer:$(sudo cat /var/lib/ubuntu-advantage/private/machine-token.json | jq -r '.resourceTokens[] | select(.type=="anbox-images").token')"
+        set -x
+        sudo microbox.amc image add jammy:android13:amd64 jammy:android13:amd64
```
morphis, line 79: Why is this needed?

.github/workflows/ci.yaml
```diff
@@ -24,6 +28,74 @@ jobs:
...
+        sudo microbox.amc image sync $(sudo microbox.amc image ls | grep jammy:android13:amd64 | grep "| container " | awk '{ print $2 }')
+
+    - name: Configure env variables (for e2e tests)
+      shell: bash
+      run: |
+        set -x
+        sudo touch .env.local
+        sudo chmod 0666 .env.local
```
morphis, line 87: See comment above, why do you need 0666 here and create as root?

.github/workflows/ci.yaml
```diff
@@ -24,6 +28,74 @@ jobs:
...
+        echo "AMS_API_CERTIFICATE=anbox-cloud.crt" >> .env.local
+        echo "AMS_API_CERTIFICATE_KEY=anbox-cloud.key" >> .env.local
+        echo "AMS_API_URL=$(sudo cat /var/snap/microbox/common/dashboard/config.yaml | grep AMS_API_URL | cut -d ' ' -f2)" >> .env.local
+        echo "ASG_API_URL=$(sudo cat /var/snap/microbox/common/dashboard/config.yaml | grep ASG_API_URL | cut -d ' ' -f2)" >> .env.local
+        echo "ASG_API_TOKEN=$(sudo cat /var/snap/microbox/common/dashboard/config.yaml | grep ASG_API_TOKEN | cut -d ' ' -f2)" >> .env.local
+
+    - name: Install Node (for e2e tests)
+      uses: actions/setup-node@v4
+      with:
+        node-version: 20
```
morphis, line 97: Why 20? Isn't 22 the latest version?

js/tests/e2e/eslint.config.mjs
```diff
@@ -0,0 +1,9 @@
+import globals from "globals";
```
morphis, line 1: Needs a copyright header

js/tests/e2e/fixtures/constants.js
```diff
@@ -0,0 +1,5 @@
+const APP_NAME = "streaming-sdk-e2e-tests";
```
morphis, line 1: Needs a copyright header

js/tests/e2e/global-setup.js
```diff
@@ -0,0 +1,112 @@
+import { chromium, expect } from "@playwright/test";
```
morphis, line 1: Needs a copyright header

js/tests/e2e/global-teardown.js
```diff
@@ -0,0 +1,22 @@
+import { expect } from "@playwright/test";
```
morphis, line 1: Needs a copyright header

js/tests/e2e/index.html
```diff
@@ -0,0 +1,77 @@
+<!doctype html>
```
morphis, line 1: Needs a copyright header

js/tests/e2e/playwright.config.js
```diff
@@ -0,0 +1,96 @@
+import { defineConfig, devices } from "@playwright/test";
```
morphis, line 1: Needs a copyright header

js/tests/e2e/playwright.config.js
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * Read environment variables from file.
+ * https://github.com/motdotla/dotenv
+ */
+// require('dotenv').config();
```
morphis, line 7: Can this be dropped if not needed?

js/tests/e2e/playwright.config.js
```diff
@@ -0,0 +1,96 @@
...
+  /* Retry on CI only */
+  retries: process.env.CI ? 2 : 0,
+  /* Opt out of parallel tests on CI. */
+  workers: 1,
+  /* Reporter to use. See https://playwright.dev/docs/test-reporters */
+  reporter: "html",
+  /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
+  use: {
+    /* Base URL to use in actions like `await page.goto('/')`. */
+    // baseURL: 'http://127.0.0.1:3000',
```
morphis, line 38: Can this be dropped if not needed?

js/tests/e2e/playwright.config.js
```diff
@@ -0,0 +1,96 @@
...
+
+    /* Test against branded browsers. */
+    // {
+    //   name: 'Microsoft Edge',
+    //   use: { ...devices['Desktop Edge'], channel: 'msedge' },
+    // },
+    // {
+    //   name: 'Google Chrome',
+    //   use: { ...devices['Desktop Chrome'], channel: 'chrome' },
+    // },
```
morphis, line 87: Can this be dropped if not needed?

js/tests/e2e/server.js
```diff
@@ -0,0 +1,232 @@
+const express = require("express");
```
morphis, line 1: Needs a copyright header

js/tests/e2e/tests/join.spec.js
```diff
@@ -0,0 +1,11 @@
+import { test, expect } from "@playwright/test";
```
morphis, line 1: Needs a copyright header

morphis (changes requested):

js/tests/e2e/tests/join.spec.js
```diff
@@ -0,0 +1,11 @@
...
+
+test("join session", async ({ page }) => {
+  await page.goto(`http://127.0.0.1:2999?sessionId=${process.env.SESSION_ID}`);
```
edlerd, line 4:
```suggestion
  await page.goto(`/?sessionId=${process.env.SESSION_ID}`);
```
We should rely on the base url in playwright config, not to repeat it in each test.

.github/workflows/ci.yaml
edlerd, line 1: Probably better to keep the e2e job in a dedicated file like and reference it like https://github.com/canonical/ams/blob/main/.github/workflows/pr.yaml

js/tests/e2e/playwright.config.js
edlerd, line 7:
```suggestion
```

js/tests/e2e/playwright.config.js
edlerd, line 38:
```suggestion
    baseURL: 'http://127.0.0.1:3000',
```
This should be set. Maybe define a constant to use here and in the teardown / setup script as well.

.github/workflows/ci.yaml
lorumic: This was imported as is from the dashboard [workflow](https://github.com/canonical/ams/blob/main/.github/workflows/ui-e2e.yaml), which @edlerd prepared. Can you please answer this question? @edlerd

.github/workflows/ci.yaml
lorumic: This was imported as is from the dashboard [workflow](https://github.com/canonical/ams/blob/main/.github/workflows/ui-e2e.yaml), which @edlerd prepared. Can you please answer this question? @edlerd

.github/workflows/ci.yaml
lorumic: This was imported as is from the dashboard [workflow](https://github.com/canonical/ams/blob/main/.github/workflows/ui-e2e.yaml), which @edlerd prepared. Can you please answer this question? @edlerd

js/tests/e2e/server.js
```diff
@@ -0,0 +1,232 @@
...
+const path = require("path");
+const axios = require("axios");
+const yaml = require("yaml");
+const compressing = require("compressing");
+require("dotenv").config({ path: ".env.local" });
+
+const { APP_NAME, IMAGE_NAME } = require("./fixtures/constants");
+
+const app = express();
+const port = 2999;
```
edlerd, line 13: I think the port should come from `.env` as well. We can add a .env file and commit it with the default port value of 2999.

Then we can use it here and construct the base url fixture for the playwright test config and wherever we need the port as well.

.github/workflows/ci.yaml
lorumic: This and the following lines make sure that the image is already active when we run the actual Playwright tests, and we don't need to download them on application creation (which would increase the test duration).

.github/workflows/ci.yaml
lorumic: Node 20 is the latest LTS as of now.

js/tests/e2e/run-tests.sh
```diff
@@ -0,0 +1,19 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+npm install
+npx playwright install --with-deps
+npx playwright test
```
edlerd, line 19: Should we run those commands in docker like we do for the unit tests?

That would unify the npm version. Currently, it relies on the host npm being available. We could then drop the install npm from the workflow.

edlerd (commented):
This looks pretty good. Some ideas and observations below.

.github/workflows/ci.yaml
edlerd: It is the driver with the least dependencies that always works. I didn't benchmark performance, we could switch this out if another one is faster.

.github/workflows/ci.yaml
edlerd: We can switch the channel to the one Anbox expects.

.github/workflows/ci.yaml
edlerd: There was an access error without this. I don't remember the exact details. You can remove those two lines to get the details.

js/tests/e2e/global-setup.js
```diff
@@ -0,0 +1,112 @@
...
+  expect(createAppResponse.status).toBe(200);
+};
+
+async function globalSetup() {
+  const hasApp = await hasTestApplication();
+  if (!hasApp) {
+    await createTestApplication();
+  }
+  const browser = await chromium.launch();
+  const page = await browser.newPage();
```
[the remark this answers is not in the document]
lorumic: It's already at the closest point to where it is used. The next line is:
```
const isAppReady = await waitForAppReady(page);
```
We need `page` there, and we want to reuse that same page for `waitForSessionActive` later on.

.github/workflows/ci.yaml
lorumic: It's much better integrated with the CI job this way. It also reuses the same logic in this step:
```
    - name: Run tests
      run: |
        ./scripts/run-tests.sh
```

js/tests/e2e/run-tests.sh
lorumic: I don't think it's possible, we need all the infrastructure (lxd, microbox, etc), not only npm, inside docker, for it to work. I'd keep this as close as the dashboard workflow if possible, keeping docker out of this.

js/tests/e2e/server.js
lorumic: Why in .env? Let's just put the port as well in the constants file and make our life simpler. It will always be used together with the base URL anyway.

js/tests/e2e/fixtures/constants.js
```diff
@@ -0,0 +1,5 @@
...
+const IMAGE_NAME = "jammy:android13:amd64";
+
+exports.APP_NAME = APP_NAME;
+exports.IMAGE_NAME = IMAGE_NAME;
```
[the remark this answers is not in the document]
lorumic: Nope, this doesn't work in CJS.

.github/workflows/ci.yaml
morphis: There is a trade-off with dir in that you need to always copy the entire rootfs or extract the rootfs tarball each time you start a new instance. In case of a loop based storage pool you can easily depend on COW but potentially have slower IO. I think it's definitely worth trying using a ZFS based pool.

js/tests/e2e/run-tests.sh
morphis: Why would you need all of LXD, microbox, etc inside docker? Isn't playwright just talking to the things over IP?

.github/workflows/ci.yaml
lorumic: That would be `5.0/stable`, right?

js/tests/e2e/run-tests.sh
lorumic: I'm afraid this goes beyond my understanding. If you have a concrete suggestion on what you want to see done here, we can try that. Otherwise this is what I could come up with with my current skills and knowledge.

lorumic: Need more info, will sync with @edlerd on this point.

edlerd: LXD and microbox are not started, they will need to accessed by IP, right. We would just run playwright wrapped in a docker container.

js/tests/e2e/server.js
edlerd: Can also be in a constants file. Though might be better to keep the config in one location, hence I was suggesting the .env file.

.github/workflows/ci.yaml
edlerd: Maybe it is better to split the unit and e2e test workflows and they can both run their entrypoint scripts. I am not too opinionated on this one though.

js/tests/e2e/fixtures/constants.js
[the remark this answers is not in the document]
edlerd: Maybe use the cjs file extension then?

js/tests/e2e/server.js
lorumic: The .env.local file contains only config related to AMS and ASG, so it would be out of place there. The constants file, instead, has the base URL, which is directly connected to the port (and needs it). I moved the server port constant there.

.github/workflows/ci.yaml
lorumic: Let's maybe hear from @morphis what he prefers here?

js/tests/e2e/run-tests.sh
lorumic: Done (thanks @edlerd )

.github/workflows/ci.yaml
```diff
@@ -24,6 +28,67 @@ jobs:
...
+        set -x
+        openssl req -x509 -newkey rsa:4096 -keyout anbox-cloud.key -out anbox-cloud.crt -days 365 -nodes -subj "/C=GB/ST=London/L=London/O=Global Security/OU=IT Department/CN=example.com"
+
+    - name: Install microbox (for e2e tests)
+      run: |
+        set -x
+        sudo pro attach ${{ secrets.UBUNTU_PRO_TOKEN }}
+        sudo pro enable --access-only anbox-cloud
+        sudo apt update
+        sudo apt install -y anbox-modules-dkms-122 linux-generic-hwe-22.04
```
adglkh, line 43: 1. Why not load the installed kernel modules after installing the anbox-modules-dkms package?
2. Why do we need to install hwe kernel? if there is a need, a reboot is required so the hwe kernel will be in use instead after that?

.github/workflows/ci.yaml
```diff
@@ -24,6 +28,67 @@ jobs:
...
+        set -x
+        sudo microbox.amc node add --unmanaged --storage-pool=local local 127.0.0.1
+        sudo cp anbox-cloud.crt /root/client.crt
+        sudo microbox.amc config trust add /root/client.crt
+        sudo microbox.amc config set registry.mode manual
+        sudo microbox.amc config set images.version_lockstep false
+        sudo microbox.amc config set images.url https://images.anbox-cloud.io/stable/
+        # NOTE We need to hide the token here to not accidentially leak credentials
+        set +x
+        sudo microbox.amc config set images.auth "bearer:$(sudo cat /var/lib/ubuntu-advantage/private/machine-token.json | jq -r '.resourceTokens[] | select(.type=="anbox-images").token')"
```
adglkh, line 75: @morphis  Is it fine to leak the way how to fetch the token that is used to access our image server?

.github/workflows/ci.yaml
```diff
@@ -24,6 +28,67 @@ jobs:
...
+        set -x
+        sudo microbox.amc image add jammy:android13:amd64 jammy:android13:amd64
+        sudo microbox.amc image sync $(sudo microbox.amc image ls | grep jammy:android13:amd64 | grep "| container " | awk '{ print $2 }')
+
+    - name: Configure env variables (for e2e tests)
+      working-directory: js/tests/e2e
+      shell: bash
+      run: |
+        set -x
+        touch .env.local
```
adglkh, line 85: Can be dropped

adglkh (changes requested):

.github/workflows/ci.yaml
lorumic: Not sure what you want to drop here?

.github/workflows/ci.yaml
lorumic: This was imported as is from the dashboard [workflow](https://github.com/canonical/ams/blob/main/.github/workflows/ui-e2e.yaml), which @edlerd prepared. Can you please answer these questions? @edlerd

.github/workflows/ci.yaml
lorumic: Ah, I see! Thanks for pointing this out. Will remove it.

ajanon (changes requested):
Some comments from me. I don't have much to say for the JS files, as I don't feel experienced enough in JS to review them.

Also, would you mind splitting this in multiple commits? Some of these changes are their own thing:
- adding a line to .gitignore
- adding the e2e directory and everything required for testing (js/tests/e2e/run-tests.sh, changes to js/run-tests.sh)
- changing the CI workflow to run them

Or maybe you wanted to wait after the first round of reviews to do that?

.github/workflows/ci.yaml
```diff
@@ -24,6 +28,66 @@ jobs:
...
+    - name: Configure env variables (for e2e tests)
+      working-directory: js/tests/e2e
+      shell: bash
+      run: |
+        set -x
+        echo "AMS_API_CERTIFICATE=anbox-cloud.crt" >> .env.local
+        echo "AMS_API_CERTIFICATE_KEY=anbox-cloud.key" >> .env.local
+        echo "AMS_API_URL=$(sudo cat /var/snap/microbox/common/dashboard/config.yaml | grep AMS_API_URL | cut -d ' ' -f2)" >> .env.local
+        echo "ASG_API_URL=$(sudo cat /var/snap/microbox/common/dashboard/config.yaml | grep ASG_API_URL | cut -d ' ' -f2)" >> .env.local
+        echo "ASG_API_TOKEN=$(sudo cat /var/snap/microbox/common/dashboard/config.yaml | grep ASG_API_TOKEN | cut -d ' ' -f2)" >> .env.local
```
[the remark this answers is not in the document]
lorumic: Yeah, let's keep things simpler without adding other packages.

.gitignore
```diff
@@ -1,2 +1,3 @@
 .npm
 node_modules
+.idea
```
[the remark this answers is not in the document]
lorumic: This was @edlerd's request. I will remove it for now, then add it at the end with a separate commit.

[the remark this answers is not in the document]
ajanon: Thanks!

ajanon (approved):
LGTM

.github/workflows/ci.yaml
morphis: There is no secret in here, so not a problem at all.

.github/workflows/ci.yaml
morphis: I think being able to run both in parallel makes sense, specifically as the e2e tests may take much longer.

.github/workflows/ci.yaml
lorumic: Removed them and there is no access error.

.github/workflows/ci.yaml
lorumic: Switching to ZFS driver for the pool. Marking as resolved.

.github/workflows/ci.yaml
lorumic: Removed these as well. Marking as resolved.

.github/workflows/ci.yaml
lorumic: Ok, I will split them.

.github/workflows/ci.yaml
lorumic: Addressed both points. Marking as resolved.

.github/workflows/ci.yaml
lorumic: I removed this now, because with the dedicated workflow the image does not become active in time, as there is no build step anymore (which used to take ~5 minutes before, allowing plenty of time for the image to be fully downloaded).

lorumic:
@morphis @adglkh @ajanon Thanks for your comments and suggestions. I think I have addressed all of them but please let me know if I forgot anything. Otherwise, this is now ready for another round of review from your side. The check will fail here as I cannot access the secret from my fork, but you can see it is now passing [here](https://github.com/lorumic/anbox-streaming-sdk/actions/runs/9283466155/job/25543740107).

@edlerd Feel free to have another pass as well, and don't worry, I'll add the `.idea` to the `.gitignore` just in time before merging - as a separate commit as @ajanon suggested. :)

adglkh (approved):
LGTM

ajanon (approved):
LGTM

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,101 @@
...
+    name: e2e-tests
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@main
+
+    - name: Store current date and time in output
+      id: datetime
+      run: echo "datetime=$(date -u +'%Y-%m-%dT%H.%M.%SZ')" >> $GITHUB_OUTPUT
+
+    - name: Generate keys and allow permissions
```
edlerd, line 32:
```suggestion
    - name: Generate keys
```

js/tests/e2e/playwright.config.js
```diff
@@ -0,0 +1,79 @@
...
+  /* Retry on CI only */
+  retries: process.env.CI ? 2 : 0,
+  /* Opt out of parallel tests on CI. */
+  workers: 1,
+  /* Reporter to use. See https://playwright.dev/docs/test-reporters */
+  reporter: "html",
+  /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
+  use: {
+    baseURL: BASE_URL,
+    ignoreHTTPSErrors: true,
```
edlerd, line 48:
```suggestion
```
We are not using ssl I think.

js/tests/e2e/playwright.config.js
```diff
@@ -0,0 +1,79 @@
...
+      use: {
+        ...devices["Desktop Firefox"],
+      },
+    },
+  ],
+
+  /* Run your local dev server before starting the tests */
+  webServer: {
+    command: "npm run start",
+    url: "http://127.0.0.1:2999",
```
edlerd, line 76:
```suggestion
    url: BASE_URL,
```
Can we use the base url here?

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,101 @@
+on:
```
edlerd, line 1:
```suggestion
name: E2e tests
on:
```
Should give this workflow a nice name

edlerd (approved):
LGTM, some tiny improvement ideas below.

js/tests/e2e/playwright.config.js
lorumic: At least when running the tests locally, for sure we are. So we need this or Playwright will get stuck when trying to join a session.

js/tests/e2e/playwright.config.js
lorumic: Yes, nice catch!

.github/workflows/e2e.yaml
lorumic: Great catch :)

.github/workflows/e2e.yaml
lorumic: Ok, wasn't sure as the one we had for the dashboard was anonymous. Will add the name there.

edlerd (approved):

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,105 @@
...
+    secrets:
+      UBUNTU_PRO_TOKEN:
+        required: true
+  workflow_dispatch:
+  schedule:
+    # execute the action nightly
+    - cron: '0 10 * * *'
+  push:
+    branches:
+      - '**'
```
morphis, line 14: Why do we run on merge (push) and nightly?

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,105 @@
...
+
+    - name: Install microbox
+      run: |
+        set -x
+        sudo pro attach ${{ secrets.UBUNTU_PRO_TOKEN }}
+        sudo pro enable --access-only anbox-cloud
+        sudo apt update
+        sudo apt install -y anbox-modules-dkms-122 linux-modules-extra-$(uname -r)
+        for n in binder_linux anbox_sync virt_wifi ; do sudo modprobe "$n" ; done
+        sudo snap install microbox --channel=latest/candidate
```
morphis, line 47: Can you add a FIXME comment and record a bug that this should switch back to latest/edge ASAP as soon as the issues are resolved?

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,105 @@
...
+        sudo microbox.amc config trust add /root/client.crt
+        sudo microbox.amc config set container.security_updates false
+        sudo microbox.amc config set registry.mode manual
+        sudo microbox.amc config set images.version_lockstep false
+        sudo microbox.amc config set images.url https://images.anbox-cloud.io/stable/
+        # NOTE We need to hide the token here to not accidentially leak credentials
+        set +x
+        sudo microbox.amc config set images.auth "bearer:$(sudo cat /var/lib/ubuntu-advantage/private/machine-token.json | jq -r '.resourceTokens[] | select(.type=="anbox-images").token')"
+        set -x
+        sudo microbox.amc image add jammy:android13:amd64 jammy:android13:amd64
```
morphis, line 79: Do we plan to run the tests for arm64 as well?

morphis (approved):
LGTM but left some questions inline

.github/workflows/e2e.yaml
lorumic: Good point - there is no specific reason actually. I'm just mimicking (= copy-pasting from) what we do for the CI build workflow. Please let me know if this makes little sense, and feel free to propose a different running conditions for the e2e tests.

.github/workflows/e2e.yaml
lorumic: Yes, I can do that. We can also wait for the fix that you've prepared for microbox, and merge this one after that (and then reverting back to edge before merging here). There's no rush. Let me know how you prefer to proceed.

.github/workflows/e2e.yaml
lorumic: I updated the job name from the (redundant) "e2e-tests" to "microbox, amd64" for better identification of the tool and architecture of the current test suite.

.github/workflows/e2e.yaml
lorumic: Added the FIXME comment :)

morphis: This should have been fixed. Can you either update the PR or send a follow up one to switch us back to latest/edge?

morphis (approved):
LGTM

morphis (approved):
LGTM

lorumic:
Cool, the workflow run passed on the [fork](https://github.com/lorumic/anbox-streaming-sdk/actions/runs/9373288727). This is ready to merge. :)

.github/workflows/e2e.yaml
lorumic: Done, thanks for the quick fix!

PR #34, opened 2024-06-06
fix(e2e-workflow): execute action on pull_request_target event

lorumic:


.github/workflows/e2e.yaml
```diff
@@ -1,18 +1,7 @@
...
-        required: true
-  workflow_dispatch:
-  schedule:
-    # execute the action nightly
-    - cron: '0 10 * * *'
-  push:
-    branches:
-      - '**'
-  pull_request:
+  pull_request_target:
```
morphis, line 4: Why don't make `e2e.yaml` only be triggered on `workflow_call` to make it reusable and have separate workflows for each trigger?

lorumic: You mean something like this:
```
on:
  workflow_call:
    secrets:
      UBUNTU_PRO_TOKEN:
        required: true
  workflow_dispatch:
  pull_request_target:
    branches:
      - '**'
```
?

adglkh: I guess Simon means 
- keep the workflow_call as is here to remain the workflow reusable
- drop the `workflow_dispatch:` here to remove the nightly build
- use e2e workflow from ci or release workflow for each pull request

adglkh: I understand your point. 

My idea here was that for the release workflow, it would trigger end-to-end (e2e) tests as well. This ensures that built assets(SDK tarball) is validated by e2e tests in addition to unit tests before a release, providing us with insights into how the change works in practice.

So the question boiled down to do we need to include e2e tests to the release workflow? if not, I'm fine you leave it as is.

Regarding your concern about the "increasing total running time," based on what I observed in https://github.com/canonical/anbox-streaming-sdk/actions, it doesn't take much longer currently. 

Generally,
- The CI job takes ~ 5 minutes.
- The e2e job takes ~ 10 minutes.

Even if we don't optimize our test setup and let both run sequentially, a 15-minute test execution time in total should be acceptable for now.

lorumic: Ok, but my main concern is that the jobs run on different conditions:
* build runs on `pull_request`
* e2e runs on `pull_request_target`

Should we make build run on `pull_request_target` as well?

lorumic: @morphis @adglkh Made the e2e job reusable and made it run, together with the build one, on `pull_request_target` event with the CI workflow. See the relevant workflow run in this PR on my fork: https://github.com/lorumic/anbox-streaming-sdk/pull/2/checks

adglkh: Thanks, that looks great to me@
One last question: can we rerun the e2e tests and run it for the release workflow as well?
@lorumic  @morphis

.github/workflows/ci.yaml
```diff
@@ -8,7 +8,7 @@ on:
   push:
     branches:
       - '**'
-  pull_request:
+  pull_request_target:
```
morphis, line 11: I don't understand how this can work. By changing to `pull_request_target` here the entire set of jobs below will run within the base of a PR and with that never test the changes introduced by a PR. @lorumic can you elaborate what this is supposed to do?

morphis (changes requested):

.github/workflows/ci.yaml
lorumic: I thought this was the approach we agreed on: https://chat.canonical.com/canonical/pl/p3xz1ajfgpy1mdj5ppj1corrxa

I will close the PR because after 2 weeks we're back to where we started and I don't understand where all this is going. I don't have other ideas to fix the issue mentioned in the thread linked above. Then, e2e tests will keep on failing to run when a PR is opened by someone from a fork, due to lack of access to the secrets.

morphis: I think what we talked about was to use the dedicated e2e workflow, switch the trigger to `pull_request_target` and then merge the branch from the PR only when someone with access rights approves (through the use of a target environment in the repository settings, see [here](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment#required-reviewers)). This here switched the entire ci workflow to run on `pull_request_target` which will render any tests we run useless as we're not testing the proposed changes. 

Any reason you dropped the idea of the target environment to have dedicated approvers before we merge with the candidate branch on the `pull_request_target` trigger?

lorumic: > Any reason you dropped the idea of the target environment to have dedicated approvers before we merge with the candidate branch on the `pull_request_target` trigger?

I never dropped that idea. That's something that only the owner of the repo can take care of - so not me. And I tried for 2 weeks to get clarifications on the concrete meaning of the first comment of this thread with no success. Without that clarification, my latest change was just an interpretation attempt.

I have created a Jira [ticket](https://warthogs.atlassian.net/browse/WD-12279) for this that anyone can pick up, as it seems that we don't have a concrete and clear idea of what to do to fix this issue.

lorumic: Just to clarify, if this is true:

> This here switched the entire ci workflow to run on `pull_request_target` which will render any tests we run useless as we're not testing the proposed changes.

(and after doing some research, it seems to be true) then this approach **will not work** as **the same applies to e2e tests**. So <ins>we're not e2e testing on the PR branch, but on the base branch</ins>, which renders the e2e tests equally useless for the same exact reason. So this approach:

> I think what we talked about was to use the dedicated e2e workflow, switch the trigger to `pull_request_target`

(which was exactly what I was doing in this PR to begin with, by the way) is to be revised completely as `pull_request_target` is not doing what we want. CC @edlerd since you're planning to work on this in the future.

edlerd: we'd use that trigger, but checkout code from the forked pr manually:

```
- uses: actions/checkout@v2
  with:
    ref: ${{github.event.pull_request.head.ref}}
    repository: ${{github.event.pull_request.head.repo.full_name}}
```

This is not advised by github, but if we put the manual approval step I think it is okay for us.

PR #35, opened 2024-06-17
fix(e2e-server): pass the zip stream directly to axios

lorumic:
Removed unnecessary conversion of the zip stream to buffer, as axios is equipped to handle a stream directly.


adglkh (approved):
LGTM

PR #37, opened 2024-06-24
fix(ci) run end to end tests on pull request target event

edlerd:
# Done
- change pr e2e trigger to fix ci

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -9,10 +9,7 @@ on:
   schedule:
     # execute the action nightly
     - cron: '0 10 * * *'
-  push:
-    branches:
-      - '**'
-  pull_request:
+  pull_request_target:
     branches:
       - '**'
 
@@ -22,10 +19,17 @@ defaults:
 
 jobs:
   e2e:
+    environment: ci-end-to-end
     name: microbox, amd64
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@main
+    - uses: actions/checkout@v4
+      with:
+        # Fetch all history for all branches and tags
+        fetch-depth: 0
+
+    - name: Check out the source branch
+      run: git checkout ${{ github.event.pull_request.head.ref }}
 
     - name: Store current date and time in output
       id: datetime


.github/workflows/e2e.yaml
```diff
@@ -25,7 +22,18 @@ jobs:
...
+        # Fetch all history for all branches and tags
+        fetch-depth: 0
+
+    - name: Set up the repository context
+      run: |
+        echo "GITHUB_BASE_REF: ${{ github.event.pull_request.base.ref }}"
+        echo "GITHUB_HEAD_REF: ${{ github.event.pull_request.head.ref }}"
+
+    - name: Check out the source branch
+      run: git checkout ${{ github.event.pull_request.head.ref }}
```
morphis, line 36: We cannot do that without approval. We *MUST* define an explicit environment the job should run on, see https://docs.github.com/en/actions/using-jobs/using-environments-for-jobs. We can then configure the environment in the repository settings to require explicit approval before this can run. Otherwise we do not really solve anything here, even worse, we now allow everyone to read our secrets :-)

morphis (changes requested):

.github/workflows/e2e.yaml
edlerd: We need to read secrets in the flow below. I thought it is okay to run this way, because we require manual approval for outside contributions in the [repository settings](https://github.com/canonical/anbox-streaming-sdk/settings/actions). So any code that is run has gone through a review beforehand.

edlerd: Or do you mean we should move the pro token to be an environment secret and use an environment here to access it?
We can do that, though that is the only secret we have at the moment.

morphis: > Or do you mean we should move the pro token to be an environment secret and use an environment here to access it?

Yes, that's what I discussed last with @lorumic from what I recall to make sure this is always explicitly approved. I am not sure if approval for outside contributors is a one-off or has to happen for every update to the PR. If it has to happen for every update to the PR this can work as well but I personally feel much better if we make the e2e test workflow optional for contributions (e.g. not have it as part of the required checks to pass) and then ask for explicit approval via an environment.

edlerd: I added the environment to the job and set up an environment secret. Hope I understood you correctly?

morphis: That looks good, thanks!

.github/workflows/e2e.yaml
```diff
@@ -22,10 +19,22 @@ defaults:
...
-    - uses: actions/checkout@main
+    - uses: actions/checkout@v4
+      with:
+        # Fetch all history for all branches and tags
+        fetch-depth: 0
+
+    - name: Set up the repository context
+      run: |
+        echo "GITHUB_BASE_REF: ${{ github.event.pull_request.base.ref }}"
+        echo "GITHUB_HEAD_REF: ${{ github.event.pull_request.head.ref }}"
```
morphis, line 34: What is this needed for? Just for debugging?

morphis (changes requested):
Overall LGTM. One small question in line. 

Also, can you update the commit message? Please drop the JIRA ticket number as that has no meaning to the public, also please put a justification/description of why the change is being implemented in the description so this can be understood by others.

.github/workflows/e2e.yaml
adglkh: I guess it just prints out branch information for the context of this workflow.
Honestly I don't see much value in keeping it here.

adglkh (approved):
LGTM. (one comment though)

.github/workflows/e2e.yaml
edlerd: Removed this step.

edlerd:
> Please drop the JIRA ticket number

This has meaning for us, as our JIRA automatically links to relevant commits and updates the ticket status on PR open (ticket is passed into review state) and on PR merge (ticket is marked as done). I kept the reference in a last line of the commit message, so it shouldn't disturb on reading the commit log.

.github/workflows/e2e.yaml
morphis, line 1: We do not see the e2e flow running anymore for this PR as we now switched to `pull_request_target` and as the workflow doesn't exist within the target repository, it is never run. So this is OK.

morphis (approved):

PR #42, opened 2024-06-26
feat(e2e): add e2e tests for stream disconnect, vhal methods [WD-11812]

lorumic:
# Done

- Determine architecture automatically by querying the `/nodes` endpoint of AMS.
- Add e2e tests for stream disconnect + `done` callback.
- Add e2e tests for VHAL: `vhalReady` callback, `getAllVhalPropConfigs`, `getVhalProperties`, `setVhalProperties` methods.

Until #37 is merged, you can see the test passing [here](https://github.com/lorumic/anbox-streaming-sdk/actions/runs/9676912820/job/26697489529).

Fixes [WD-11812](https://warthogs.atlassian.net/browse/WD-11812)

[WD-11812]: https://warthogs.atlassian.net/browse/WD-11812?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -74,6 +74,7 @@ jobs:
         set +x
         sudo microbox.amc config set images.auth "bearer:$(sudo cat /var/lib/ubuntu-advantage/private/machine-token.json | jq -r '.resourceTokens[] | select(.type=="anbox-images").token')"
         set -x
+        sudo microbox.amc image add jammy:android13:amd64 jammy:android13:amd64
         sudo microbox.amc image add jammy:aaos13:amd64 jammy:aaos13:amd64
 
     - name: Configure env variables
@@ -81,7 +82,6 @@ jobs:
       run: |
         set -x
         echo "CI=true" >> .env.local
-        echo "ARCHITECTURE=amd64" >> .env.local
         echo "AMS_API_CERTIFICATE=anbox-cloud.crt" >> .env.local
         echo "AMS_API_CERTIFICATE_KEY=anbox-cloud.key" >> .env.local
         echo "AMS_API_URL=$(sudo cat /var/snap/microbox/common/dashboard/config.yaml | grep AMS_API_URL | cut -d ' ' -f2)" >> .env.local
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -74,7 +74,7 @@ jobs:
         set +x
         sudo microbox.amc config set images.auth "bearer:$(sudo cat /var/lib/ubuntu-advantage/private/machine-token.json | jq -r '.resourceTokens[] | select(.type=="anbox-images").token')"
         set -x
-        sudo microbox.amc image add jammy:android13:amd64 jammy:android13:amd64
+        sudo microbox.amc image add jammy:aaos13:amd64 jammy:aaos13:amd64
 
     - name: Configure env variables
       shell: bash
--- a/js/tests/e2e/tests/vhal.spec.js
+++ b/js/tests/e2e/tests/vhal.spec.js
@@ -1,0 +1,95 @@
+/*
+ * This file is part of Anbox Cloud Streaming SDK
+ *
+ * Copyright 2024 Canonical Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { expect, test } from "@playwright/test";
+import { joinSession, disconnectStream } from "./shared";
+
+const VALUES = {
+  287310858: { expected: 0, target: 1 }, // ABS_ACTIVE
+  289475073: { expected: "6,0", target: "6,1" }, // AP_POWER_STATE_REPORT
+  289409539: { expected: 82, target: 100 }, // DISPLAY_BRIGHTNESS
+  291504900: { expected: 101, target: 101.5 }, // ENGINE_OIL_TEMP
+  286261505: { expected: "Toy Vehicle", target: "Foo Bar" }, // INFO_MAKE
+};
+
+const TARGET_VHAL_PROPS = Object.keys(VALUES).map((id) => +id);
+
+test("read VHAL prop configs, get VHAL prop values, set VHAL prop values", async ({
+  page,
+}) => {
+  await joinSession(page);
+
+  await page.waitForFunction(
+    () =>
+      globalThis.vhalPropConfigs !== undefined &&
+      globalThis.vhalPropConfigs.length > 0,
+    null,
+    {
+      timeout: 20_000,
+    },
+  );
+
+  const vhalProperties = await page.evaluate(async (TARGET_VHAL_PROPS) => {
+    const filteredPropConfigs = globalThis.vhalPropConfigs.filter((prop) =>
+      TARGET_VHAL_PROPS.includes(prop.prop),
+    );
+    const vhalProps =
+      await globalThis.stream.getVhalProperties(filteredPropConfigs);
+    return vhalProps;
+  }, TARGET_VHAL_PROPS);
+
+  expect(vhalProperties.length).toBe(TARGET_VHAL_PROPS.length);
+  expect(vhalProperties.some((getResult) => getResult.error)).toBe(false);
+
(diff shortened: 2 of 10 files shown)


ajanon (commented):
Just one small question: with this change, we would only be running e2e tests against AAOS 13, right?
Should we run these tests against multiple LXD images versions?
I'm not sure if it's too useful, but I would like to have other opinions.

lorumic:
> Just one small question: with this change, we would only be running e2e tests against AAOS 13, right? Should we run these tests against multiple LXD images versions? I'm not sure if it's too useful, but I would like to have other opinions.

For now, I'm focusing on having a single image downloaded, and running tests on a session created from that one only. This is to make the tests run reasonably fast. If there is an actual need of running on different images, technically we could do that, however I don't know if it's in the scope of the e2e tests, as it seems more about "testing the images" rather than "testing the features of the Anbox streaming SDK end-to-end". However, this is just my point of view, so I'm happy to hear other opinions and change the workflow accordingly. :)

ajanon (approved):
LGTM then, thanks!

morphis:
> For now, I'm focusing on having a single image downloaded, and running tests on a session created from that one only. This is to make the tests run reasonably fast. If there is an actual need of running on different images, technically we could do that, however I don't know if it's in the scope of the e2e tests, as it seems more about "testing the images" rather than "testing the features of the Anbox streaming SDK end-to-end". However, this is just my point of view, so I'm happy to hear other opinions and change the workflow accordingly. :)

There are quite a few nuances we do not necessarily can test with every image (e.g. I doubt we have virtual keyboard support for AAOS atm). I think you definitely want one per image variant (android, aaos and generic as well once we start officially supporting that).

.github/workflows/e2e.yaml
```diff
@@ -74,7 +74,7 @@ jobs:
         set +x
         sudo microbox.amc config set images.auth "bearer:$(sudo cat /var/lib/ubuntu-advantage/private/machine-token.json | jq -r '.resourceTokens[] | select(.type=="anbox-images").token')"
         set -x
-        sudo microbox.amc image add jammy:android13:amd64 jammy:android13:amd64
+        sudo microbox.amc image add jammy:aaos13:amd64 jammy:aaos13:amd64
```
morphis, line 77: See comment on the PR. I think we start switching the image we should run the same tests for both AAOS and Android as there are different nuances we want to test as well going forward

js/tests/e2e/tests/vhal.spec.js
```diff
@@ -0,0 +1,95 @@
...
+};
+
+const TARGET_VHAL_PROPS = Object.keys(VALUES).map((id) => +id);
+
+test("read VHAL prop configs, get VHAL prop values, set VHAL prop values", async ({
+  page,
+}) => {
+  await joinSession(page);
+
+  await page.waitForFunction(
```
morphis, line 37: What about `isVhalAvailable`?

It would also be nice to test the opposite case where we do not have the VHAL (e.g. with `jammy:android12`) and make sure things are correctly reported. Given that is currently what is deployed the most by our users I think that definitely deserves a test case. However you may want to split this in a future story if it expands the scope of what you're currently working on.

morphis (changes requested):

morphis:
@lorumic I've merged https://github.com/canonical/anbox-streaming-sdk/pull/37, if you update the PR we should be able to get tests running now.

js/tests/e2e/tests/vhal.spec.js
lorumic: > What about `isVhalAvailable`?

No need to check for that when you have the `vhalReady` callback - which is anyway a better and more reliable way to test for VHAL capability asynchronously.

> It would also be nice to test the opposite case where we do not have the VHAL (e.g. with `jammy:android12`) and make sure things are correctly reported. Given that is currently what is deployed the most by our users I think that definitely deserves a test case. However you may want to split this in a future story if it expands the scope of what you're currently working on.

What does this mean? Test that `isVhalAvailable()` returns `false` when using an AOSP image? That's tricky because it will always be false at the beginning, but this doesn't necessarily mean that it never becomes true. Or are you suggesting to wait for a given time to check that it doesn't become true, and if it doesn't then we're good?

.github/workflows/e2e.yaml
lorumic: For now, we're not testing those nuances, are we? When we'll do that, we can consider downloading different images, increase the timeouts so that the download can finish, and test those nuances.

But for now, that would only mean unnecessarily increase the test run time for no added benefit, since what we're testing is 100% equivalent on AAOS, AOSP and probably even generic.

lorumic:
@morphis @ajanon 

I'm laying the ground for when we'll need to test different features on different images. For this purpose, I tested locally the parallel creation of an AOSP-based app + session and AAOS-based ones. Thanks to the parallel execution, it should not take too much additional time - the test is still running, though, and we'll have to see how it behaves on the runner, since we'd also need to download both images there, unlike locally where they are already downloaded.

If that works well, I will run the join + disconnect stream test on the AOSP session, and the VHAL test (which includes join + disconnect anyway) on the AAOS one. Anything else, I'd like to keep as a future addition so as not to grow too much the scope of this task, which was planned to comprise only what is listed in the "Done" section of the PR description above.

Does that sound reasonable to you?

js/tests/e2e/tests/vhal.spec.js
```diff
@@ -0,0 +1,95 @@
...
+  setInputs.push(getSetInput(289409539, "int32_values")); // DISPLAY_BRIGHTNESS
+  setInputs.push(getSetInput(291504900, "float_values")); // ENGINE_OIL_TEMP
+  setInputs.push(getSetInput(286261505, "string_value")); // INFO_MAKE
+
+  const setProperties = await page.evaluate(async (setInputs) => {
+    const res = await globalThis.stream.setVhalProperties(setInputs);
+    return res;
+  }, setInputs);
+
+  expect(setProperties.length).toBe(TARGET_VHAL_PROPS.length);
```
adglkh, line 91: Can we read and verify the values again after modifying them?

.github/workflows/e2e.yaml
lorumic: Thanks for your comment Gary. I have partly implemented the changes requested by @ajanon and @morphis, see [this](https://github.com/canonical/anbox-streaming-sdk/pull/42#issuecomment-2191527283) comment. This allows to test both AOSP and AAOS images (on Android 13 for now, but please do let me know if you prefer e.g. AOSP 12 and I can easily change it in the workflow). I think this can be a good start, and the overhead in terms of test run time is just +3 minutes.

js/tests/e2e/tests/vhal.spec.js
lorumic: Thanks for the suggestion @adglkh, the check has now been implemented. Test run: https://github.com/lorumic/anbox-streaming-sdk/actions/runs/9681782608

js/tests/e2e/tests/vhal.spec.js
morphis: The main part we need to cover is that in case that we do not expect any VHAL capabilities from the image we should get sane results from all the API calls the SDK has. That starts with `isVhalAvailable` but applies for the other methods as well.

morphis: > No need to check for that when you have the vhalReady callback - which is anyway a better and more reliable way to test for VHAL capability asynchronously.

I agree. However we should still check that `isVhalAvailable` gives us the expected value at that point. It's a public API method so to ensure we have full coverage we should call it and match against our expectation.

morphis (changes requested):

js/tests/e2e/tests/vhal.spec.js
```diff
@@ -0,0 +1,186 @@
...
+    {},
+    ...vhalPropsBefore.map(getValueObject),
+  );
+  const valuesAfter = Object.assign({}, ...vhalPropsAfter.map(getValueObject));
+  checkChanges(valuesBefore, setInputs, valuesAfter);
+
+  await disconnectStream(page);
+});
+
+test("VHAL methods on an AOSP session", async ({ page }) => {
```
morphis, line 177: Nice, thanks a lot for doing this!

morphis (approved):
LGTM

adglkh (approved):
LGTM

ajanon (approved):
LGTM

morphis:
@lorumic I think we're blocked on https://github.com/canonical/anbox-streaming-sdk/pull/43 before we can merge this one?

lorumic:
As discussed with @edlerd, there are changes to the workflow in this PR which won't get picked up, making the tests still fail. I have another addition of e2e tests in the works (based on this one) which doesn't make changes to the workflow, so we can test the CI changes on that one once I open it. You can check that the tests for this branch are passing by looking at the fork [run](https://github.com/lorumic/anbox-streaming-sdk/actions/runs/9757080832/job/26928703020#step:10:1882). So I'm merging this now.

PR #43, opened 2024-06-26
fix(ci) ensure a branch from a forked pr is available in ci

edlerd:
We would previously checkout the canonical repo, but for a PR from a
 fork we need to checkout the fork repo instead.
 WD-12279

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -25,11 +25,7 @@ jobs:
     steps:
     - uses: actions/checkout@v4
       with:
-        # Fetch all history for all branches and tags
-        fetch-depth: 0
-
-    - name: Check out the source branch
-      run: git checkout ${{ github.event.pull_request.head.ref }}
+        ref: "${{ github.event.pull_request.merge_commit_sha }}"
 
     - name: Store current date and time in output
       id: datetime


.github/workflows/e2e.yaml
```diff
@@ -25,11 +25,8 @@ jobs:
     steps:
     - uses: actions/checkout@v4
       with:
-        # Fetch all history for all branches and tags
-        fetch-depth: 0
-
-    - name: Check out the source branch
-      run: git checkout ${{ github.event.pull_request.head.ref }}
+       ref: ${{ github.event.pull_request.head.ref }}
+       repository: ${{ github.event.pull_request.head.repo.full_name }}
```
morphis, line 29: Doesn't this mean we run on what the latest state in the remote repository is? Github actually does a merge with the target repository for the checkout you test in so you run a combined state in case that the PR is not up-to-date. See the following from the CI execution of this PR:

image

Specifically the last line is important: `HEAD is now at e2b8467 Merge 9b296ed4db32c448e46a14055e6740051bb1c457 into b63a98e8b7607e7dbcbb5d24cc9d04c393292088`

It checks out commit e2b8467 whereas 9b296ed4db32c448e46a14055e6740051bb1c457 is top of your to be merged branch and b63a98e8b7607e7dbcbb5d24cc9d04c393292088 is what is latest in the repository on the main branch. With this change here we will checkout 9b296ed4db32c448e46a14055e6740051bb1c457 and ignore any changes in main.

morphis (changes requested):

.github/workflows/e2e.yaml
edlerd: If found [this](https://github.com/actions/checkout/issues/518#issuecomment-890401887), which suggests relying on `ref: "${{ github.event.pull_request.merge_commit_sha }}"`, because there is a race when a run from outside collaborators get approved. The external PR might be modified between approval and the GH action doing the checkout of the external branch. Relying on the merge sha prevents this race attack.

edlerd: Changed this to `ref: "${{ github.event.pull_request.merge_commit_sha }}"`, which is the merge hash on the target repo for the open PR. this is being discussed in [this thread](https://github.com/actions/checkout/issues/518), with some concerns the sha might be out of date. I think in that case, we'd need to keep in mind to trigger CI a 2nd time.

morphis: @edlerd what about doing the merge ourself? We have the candidate branch and know the target. Doing a `git merge ...` should be straight forward if we start with a target checkout and merge the candidate branch in?

edlerd: I agree, we can bundle our own merge strategy in the workflow. The commit sha is the only reliable way to ensure we run exactly the code that was reviewed by us. Relying on the PR branch name opens an attack vector to run code that was not reviewed. The user can push to their fork-branch after things got reviewed by us and before the workflow is running.

morphis (approved):
LGTM

.github/workflows/e2e.yaml
morphis: Thanks for the explanation @edlerd !

PR #45, opened 2024-07-02
feat(tests): visual tests for input events: touch, keypress, long press, swipe [WD-12195]

lorumic:
# Done

Added visual tests for input / interaction events:

- Simple touch event ("tap")
- Keyboard input
- Long press
- Swipe gesture

Fixes [WD-12195](https://warthogs.atlassian.net/browse/WD-12195)

[WD-12195]: https://warthogs.atlassian.net/browse/WD-12195?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


adglkh (changes requested):
Can we create symbolic links pointing to the Chromium screenshots when Firefox is in use, given that they are pretty much the same?

lorumic:
> Can we create symbolic links pointing to the Chromium screenshots when Firefox is in use, given that they are pretty much the same?

@adglkh Thanks for the suggestion. It totally makes sense, and I have done that. Please note that for one of them we cannot do it, as for some reason it has a small difference between Chromium and Firefox by default (the microphone button on the right of the search bar).

adglkh (approved):
LGTM

js/tests/e2e/playwright.config.js
```diff
@@ -60,13 +60,20 @@ export default defineConfig({
...
     },
 
     {
       name: "firefox",
       use: {
         ...devices["Desktop Firefox"],
+        ...SHARED_BROWSER_OPTIONS,
+        launchOptions: {
+          firefoxUserPrefs: {
+            "media.videocontrols.picture-in-picture.video-toggle.enabled": false,
```
edlerd, line 74: Can you add a comment why we disable this?

js/tests/e2e/tests/visual.spec.js-snapshots/text-input-firefox-linux.png
```diff
@@ -0,0 +1 @@
+text-input-chromium-linux.png
```
edlerd, line 1: Is there a way to share the screenshots without having the dedicated files for firefox and chromium? It seems they are all the same as the ff screens reference the chromium screenshots.

lorumic: The screenshot filename follows a specific pattern that Playwright handles and expects to find:

`screenshot-name-browser-osname`

Where only the `screenshot-name` is the part that you're allowed to customise, i.e. the string that you pass as a parameter of the `toHaveScreenshot` function (plus the extension). The browser name and OS name (linux, darwin, etc) are managed by Playwright and cannot be customised.

js/tests/e2e/playwright.config.js
lorumic: Sure thing, that's done

edlerd (approved):
LGTM, thanks for adding those :+1:

PR #48, opened 2024-08-19
feat(gh): drop secret use from e2e workflow

morphis:
We had to do quite a dance to get tests working as the e2e test workflow requires access to the Ubuntu Pro secret. We need to be careful to not allow access to it on a PR from a fork. The approach so far by using the pull_request_target trigger and a dedicated deployment environment with approval was working but far from perfect.

This now implements an alternative which does not require the direct use of the Ubuntu Pro token in the test workflow. Instead we use it to cache images in an independent workflow. The workflow will download all needed images and store them in the Github Action cache under a well known key name. When the e2e workflow now runs it will fetch the images from the cache and load into AMS. The workflow runs every day so we keep the images fresh.

-----

**NOTE** As the cache workflow still has to run the initial execution on the PR itself will fail.

### DIFF
--- a/.github/workflows/cache-images.yaml
+++ b/.github/workflows/cache-images.yaml
@@ -1,0 +1,35 @@
+name: Cache Anbox images
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 10 * * *'
+
+jobs:
+  cache-images:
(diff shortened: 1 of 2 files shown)


ajanon (approved):
LGTM

PR #49, opened 2024-08-19
feat(gh): update renovate configration

morphis:
This matches what we implement elsewhere, specifically it pins actions now to specific digests rather than tags.


ajanon (approved):
LGTM

PR #50, opened 2024-08-19
Various cleanups for the different workflows we run

morphis:
See commit messages for details.


ajanon (approved):
LGTM

js/tests/e2e/global-setup.js
```diff
@@ -144,11 +144,11 @@ const setupSessionFor = async (page, appName) => {
...
-  const sessionIds = await Promise.all(
-    [AOSP_APP_NAME, AAOS_APP_NAME].map((appName) => {
-      return setupSessionFor(page, appName);
-    }),
-  );
+  var sessionIds = [];
+  for (const appName of [AOSP_APP_NAME, AAOS_APP_NAME]) {
+    const id = await setupSessionFor(page, appName);
+    sessionIds.push(id);
+  }
```
edlerd, line 151: Not sure if this is better. It has potential to take much longer, if the session setup is taking a long time twice. This will then make the suite take longer. Do we have reason to believe this step is faster now, sequentially?

edlerd (commented):
LGTM, one question on the session setup change

js/tests/e2e/global-setup.js
lorumic: It will definitely not be. Please revert this change.

lorumic (changes requested):

js/tests/e2e/global-setup.js
lorumic: Also, a JS array does not need to be declared as `var` when you only need to push items to it. It can be `const`.

morphis: Have a look at the [commit message](https://github.com/canonical/anbox-streaming-sdk/pull/50/commits/e1eb317ed377d5b0f1c288f444b210fa1c1a8a26) for details. The reason is not this is faster but it takes less resources and provides high stability. I have been seeing various timeouts being caused on the low constraint runner by bootstrapping two applications in parallel while putting this in place. Put bootstrap into a sequence makes this much better. 

> Also, a JS array does not need to be declared as var when you only need to push items to it. It can be const

You can modify an array declared as `const`? That is pretty weird.

aaryanporwal: > You can modify an array declared as const? That is pretty weird.
Yes, `const` is a reference to a value. We cannot change constant "primitive" values, but we can change the properties of constant objects.
`const` has no hoisiting and has block scope that's why it's preferred mode for us.

lorumic: Ok, provided that we know that this is slower, if there is another reason for it then I'm ok with the change. But please declare the array as `const`. You're not changing the reference of the array, so it has no reason to be a `var`.

lorumic (approved):

PR #51, opened 2024-09-02
Fix the canvas rotation and avoid a busy loop after stopping the streaming when upscaling is enabled

adglkh:
- Rotate the canvas rather than video element when upscaling is enabled
- Avoid a busy loop after stopping the streaming when upscaling is enabled

### DIFF
--- a/js/tests/unit/sdk.test.js
+++ b/js/tests/unit/sdk.test.js
@@ -29,6 +29,11 @@ beforeEach(() => {
     document.webkitExitFullscreen = () => {
         document.fullscreenElement = null
     }
+    // Mock the getContext function to HTMLCanvasElement
+    window.HTMLCanvasElement.prototype.getContext = () => {
+      return {};
+    };
+
     const container = document.createElement('div')
     container.__defineGetter__('clientWidth', () => { return 100 });
     container.__defineGetter__('clientHeight', () => { return 100 });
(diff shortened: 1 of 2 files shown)


js/tests/unit/sdk.test.js
```diff
@@ -196,29 +205,50 @@ test('video element should take all available space', () => {
...
+      expect(stream._webrtcManager.sendControlMessage.mock.calls[0][1]).toEqual({orientation: 'landscape'})
+      expect(stream.getCurrentOrientation()).toEqual('landscape')
 
-    expect(() => stream.disconnect()).not.toThrow()
+      expect(() => stream.disconnect()).not.toThrow()
+    };
+
+    test_video_rotation(sdkOptions)
+
+    sdkOptions.experimental.upscaling.enabled=true;
```
morphis, line 250: Looks like we miss a style checker:

```suggestion
    sdkOptions.experimental.upscaling.enabled = true;
```

morphis (changes requested):

js/tests/unit/sdk.test.js
adglkh: We only run ESLint against the anbox-stream-sdk.js file, not the test files under js/tests/unit:
- callbacks.test.js
- sdk.test.js
- input-events.test.js
- stats.test.js

Yes, it would be nice if we could enable ESLint for those files as well. 
Since it's gonna be the first time running ESLint for the unit test files, the change to those test file might be big. Hence let's not mix up the scope of this PR.
If you give a +1, I can send another PR to do this.  How does that sound?

morphis (approved):

PR #56, opened 2024-09-06
[Backport] Fix the canvas rotation and avoid a busy loop after stopping the streaming when upscaling is enabled

adglkh:
- fix(js-sdk): rotate the canvas rather than video element when upscaling is enabled
- fix(js-sdk): avoid a busy loop even after stopping the streaming
- fix(js-sdk): cancel video frame callback when video stream stops


edlerd (approved):
LGTM, in the dashboard, we do not use the rotate function.

jat-canonical (approved):
LGTM

PR #59, opened 2024-09-14
Add and integrate Snapdragon Game Super Resolution(SGSR) into js sdk

adglkh:
Snapdragon™ Game Super Resolution(SGSR), which integrates upscaling
and sharpening in one single GPU shader pass. The algorithm uses a
12-tap Lanczos-like scaling filter and adaptive sharpening filter,
which presents smooth images and sharp edges. For more details:
    
     https://github.com/SnapdragonStudios/snapdragon-gsr
    
This change adds and integrates SGSR, this single GPU shader pass
into the streaming js sdk, and usd by default for the time being.
    
NOTE: for now, FSR is still kept in the source for reference.
In the end, we will choose the best upscaling algorithm from all
available options after benchmarking and comparing performance and
quality, and only ship the best one in the streaming SDK.

### DIFF
--- a/examples/js/sgsr.html
+++ b/examples/js/sgsr.html
@@ -8,7 +8,7 @@
 
 <body>
 <script type="module">
-    import {AnboxStream, AnboxStreamGatewayConnector} from './anbox-stream-sdk.js';
+    import {AnboxStream, AnboxStreamGatewayConnector} from '../anbox-stream-sdk.js';
 
     // Connectors will enable communication between the client and the
     // gateway. They can be customized to add an additional layer between
--- a/examples/js/sgsr.html
+++ b/examples/js/sgsr.html
@@ -1,0 +1,221 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta content="text/html;charset=utf-8" http-equiv="Content-Type">
+    <meta content="utf-8" http-equiv="encoding">
+    <title>Anbox Streaming SDK Example - Enable video upscaling with Snapdragon™ Game Super Resolution(SGSR) </title>
+</head>
+
+<body>
+<script type="module">
+    import {AnboxStream, AnboxStreamGatewayConnector} from './anbox-stream-sdk.js';
+
+    // Connectors will enable communication between the client and the
+    // gateway. They can be customized to add an additional layer between
+    // the client and the gateway in order to add more features (user management,
+    // limits, analytics, etc).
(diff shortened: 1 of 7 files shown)


adglkh:
Let's hold off on merging this until we've discussed how to manage multiple shaders in the main branch or decide whether to keep FSR before we reach a conclusion after benchmarking and comparing the performance and quality of all available upscaling algorithms (FSR, Anime4K, SGSR), choose the best one, and use it in our JS SDK.

morphis (changes requested):
Can we split out SGSR integration into an example? That way we can merge and then defer the decision to a later point. Wdyt?

morphis:
@adglkh we don't want to switch the actual shader implementation in the SDK yet. So what we can do instead is write an example application using SGSR and add that to https://github.com/canonical/anbox-streaming-sdk/tree/main/examples/js as part of this PR. Then we have an example demonstrating the use and have no pending PR.

adglkh:
> @adglkh we don't want to switch the actual shader implementation in the SDK yet. So what we can do instead is write an example application using SGSR and add that to https://github.com/canonical/anbox-streaming-sdk/tree/main/examples/js as part of this PR. Then we have an example demonstrating the use and have no pending PR.

A: okay, that makes sense to me.

adglkh:
The image is broken, which has blocked e2e testing. Let's wait for the new image to be built and cached today in the next run.

morphis (approved):
LGTM

examples/js/sgsr.html
```diff
@@ -0,0 +1,218 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta content="text/html;charset=utf-8" http-equiv="Content-Type">
+    <meta content="utf-8" http-equiv="encoding">
+    <title>Anbox Streaming SDK Example - Enable video upscaling with Snapdragon™ Game Super Resolution(SGSR) </title>
```
ajanon, line 6: Random question: do we need an agreement from them to use this name?
Or put a disclaimer (e.g. 'This trademark is possessed by...')

ajanon (approved):
LGTM, just one question

examples/js/sgsr.html
adglkh: Thanks @ajanon , would you mind taking another look?

ajanon (approved):
LGTM

PR #67, opened 2024-10-22
fix(e2e): use AMS to create a streamable instance instead of ASG

lorumic:


js/tests/e2e/global-setup.js
```diff
@@ -34,35 +34,42 @@ const isTestAppAvailable = async (appName) => {
...
+  const instances = await response.json();
+  return instances.find((instance) => instance.app_name === appName);
 };
 
-const hasActiveTestSession = async (appName) => {
-  const response = await fetch(`${BASE_URL}/sessions`);
-  const sessions = await response.json();
-  const session = sessions.find(
-    (session) => session.app === appName && session.status === "active",
+const getRunningSessionId = async (appName) => {
```
edlerd, line 43:
```suggestion
const getRunningInstanceId = async (appName) => {
```
Should this name be aligned as well?

lorumic: No, this method returns a session id, not an instance id.

PR #75, opened 2024-10-30
fix(e2e): add small delay between text input and snapshot check

lorumic:
Added a 1s timeout after text input and before checking the snapshot, to reduce the flakiness of the keyboard input visual test.


adglkh:
LGTM

anusha-c18 (approved):
LGTM 👍

PR #76, opened 2024-11-01
Add the latest link for streaming tutorial

keirthana:
The streaming tutorial link will become live once https://github.com/canonical/anbox-cloud-docs/pull/205 is merged

### DIFF
--- a/js/anbox-stream-sdk.js
+++ b/js/anbox-stream-sdk.js
@@ -762,7 +762,7 @@ class AnboxStream {
       );
     } else if (container.clientWidth == 0 || container.clientHeight == 0)
       console.error(
-        "AnboxStream: video container element misses size. Please see https://anbox-cloud.io/docs/howto/stream/web-client"
+        "AnboxStream: video container element misses size. Please see https://documentation.ubuntu.com/anbox-cloud/en/latest/tutorial/stream-client"
       );
 
     if (this._nullOrUndef(options.connector)) {


adglkh (approved):
LGTM

js/anbox-stream-sdk.js
```diff
@@ -762,7 +762,7 @@ class AnboxStream {
       );
     } else if (container.clientWidth == 0 || container.clientHeight == 0)
       console.error(
-        "AnboxStream: video container element misses size. Please see https://anbox-cloud.io/docs/howto/stream/web-client"
+        "AnboxStream: video container element misses size. Please see https://documentation.ubuntu.com/anbox-cloud/en/latest/tutorial/stream-client"
```
lorumic, line 765: There are a bunch of other links using the old `anbox-cloud.io/docs` pattern, so I would rather like to see them updated all at once, for consistency, than only this one.

lorumic: @keirthana Adding to the point above - this change is currently causing a unit test to fail, because that test is still looking for the old string (i.e. the one containing `anbox-cloud.io/docs/...`, and cannot find it anymore.

lorumic (changes requested):

js/anbox-stream-sdk.js
keirthana: Okay, I created a separate PR for the other links because this one is tied to a bug. For the unit test failure, I might need some time to figure out how to fix that.

lorumic: You just need to make the text there match the text here. The test is doing string matching.
So you can simply replace this at lines 764-766 of the `js/anbox-stream-sdk.js` file:
```
      console.error(
        "AnboxStream: video container element misses size. Please see https://documentation.ubuntu.com/anbox-cloud/en/latest/tutorial/stream-client"
      );
```

lorumic: You just need to make the text there match the text here. The test is doing string matching.
So you can simply replace the lines 154-156 of the `js/tests/unit/sdk.test.js` file with this:
```
    expect.stringContaining(
      "AnboxStream: video container element misses size. Please see https://documentation.ubuntu.com/anbox-cloud/en/latest/tutorial/stream-client"
    )
```
I see that you have added this change in the other PR, but it belongs to this one. These 2 changes must go together, or the check won't pass.

keirthana: oh I understand now. I will fix that, thank you for explaining.

keirthana:
Closing this PR as this is targeting an incorrect remote branch. Created PR #78 instead

PR #81, opened 2024-11-05
feat(e2e-tests): add coverage for e2e tests

lorumic:


edlerd (approved):
LGTM

js/tests/e2e/fixtures/coverage.js
```diff
@@ -0,0 +1,69 @@
...
+  });
+};
+
+export const finishCoverage = async (page) => {
+  const coverage = await page.coverage.stopJSCoverage();
+  for (const entry of coverage) {
+    const fileMatcher = entry.url.match(/http(s)*:\/\/.*:2999\/(?<file>.*)/);
+    if (
+      !fileMatcher?.groups ||
+      fileMatcher.groups.file !== "anbox-stream-sdk.js"
```
edlerd, line 37: Isn't this the only file we are interested in for the coverage? Or does it appear with another name a second time?

lorumic: Yeah, this addition is just to filter out the index html file, otherwise that gets added to the coverage as well, and we probably don't need that.

edlerd: Oh I missed this is not equals and assumed it is equals.

edlerd (approved):

PR #94, opened 2024-12-12
fix(e2e): set session offline to avoid results in search widget

lorumic:
Fix for the failure in the "Run code quality scan" job run [here](https://github.com/canonical/anbox-streaming-sdk/actions/runs/12274448590/job/34247382351).

We are getting unexpected search results when testing the visual output of sending keyboard input to the stream (see screenshots below). It should be possible to prevent that by putting the Android instance offline.

- Expected:

image

- Received:

image


morphis (approved):
LGTM

PR #106, opened 2025-01-13
fix(stream-alignment): remove option to change vertical alignment

lorumic:
The option to change the vertical alignment of the stream is causing troubles with user interaction (click/touch), so it is removed by this PR (which effectively reverts #28).
The plan is to handle the vertical alignment directly in the front-end application where there is such a need - i.e. the AMS one.


lorumic:
Closing for now, as the deprecation policy would force this broken feature to be still kept available for 6 months, so it would probably be better (and easier) to fix the issue in the meantime. Bug opened and tracked [here](https://bugs.launchpad.net/anbox-cloud/+bug/2093904).

PR #110, opened 2025-01-31
fix(gamepad): query gamepad events when gamepad is already connected

lorumic:
## Done

- Added a call to the `_queryGamePadEvents` method also in the context of the `connect` method of the `AnboxStream`, so that we don't miss any gamepad that was connected already before we joined the session.

## QA

- Join a session stream for an Android application that can handle gamepad events (e.g. a game)
- Connect a gamepad/detect a gamepad by clicking a button/move axis on it
- See that the gamepad is working on the Android stream
- In the context of the same window/tab, navigate to another page (e.g. if you're using the Anbox Cloud dashboard, to the "Instances" page)
- Still in the context of the same window/tab, rejoin the session stream
- Ensure that the gamepad is working after rejoining

## JIRA / Launchpad bug

Fixes [LP#2097084](https://bugs.launchpad.net/anbox-cloud/+bug/2097084)


morphis (approved):
LGTM, thanks!

PR #113, opened 2025-02-10
fix(gamepad): stop polling in game manager to avoid redundant events

lorumic:
## Done

- Added a check before the initialisation of a `_gamepadEventManager` instance. If it was already initialised before, it should `stopPolling` before initialising a new one, to avoid receiving redundant gamepad events.


edlerd (approved):
LGTM, thanks for the improvement.

anusha-c18 (approved):
LGTM!

PR #120, opened 2025-02-21
Release workflow changes

ajanon:
## Done

Small changes to the release workflow: instead of tagging, and then creating the release and testing, we instead depend first build and test before tagging. We're thus not using the version generation script anymore, and instead rely on the tag name. In practice, this shouldn't change anything: if we're based on a tag, the generation script just returns the tag name.

This should avoid creating a tag, and then failing to publish the release as the tests or build failed (for instance).

Also add a way to specify a different commit than the branch head. This should give more flexibility in choosing the commit for the release, to ensure it is the aligned with the one used by the dashboard (as this is the version that was tested the most).

Last, this fills the body of the release with a link to the release notes in our docs website.

## QA

N/A

## JIRA / Launchpad bug

Contributes to AC-3134

## Documentation

Release documentation will be updated accordingly.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -18,10 +18,6 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
-      with:
-        # to avoid shadow clone which would impact the version generation
-        # See `scripts/gen-version.sh`
-        fetch-depth: "0"
 
     - name: Build
       run: |
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -16,17 +16,6 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - name: Create tag
-      uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7
-      with:
-        script: |
-          await github.rest.git.createRef({
-            owner: context.repo.owner,
-            repo: context.repo.repo,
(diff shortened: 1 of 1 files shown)


adglkh (commented):
To be confirmed, with this change, in the next release, this will no longer to create a tag in each release, is that correct?

ajanon:
> To be confirmed, with this change, in the next release, this will no longer to create a tag in each release, is that correct?

This should still create a tag.
See: https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28
```
Body parameters
`tag_name` string Required
The name of the tag.


`target_commitish` string
Specifies the commitish value that determines where the Git tag is created from. Can be any branch or commit SHA. Unused if the Git tag already exists. Default: the repository's default branch.
```

AFAIU, this means that the create release REST API creates a tag. Wdyt?

adglkh:
> > To be confirmed, with this change, in the next release, this will no longer to create a tag in each release, is that correct?
> 
> This should still create a tag. See: https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28
> 
> ```
> Body parameters
> `tag_name` string Required
> The name of the tag.
> 
> 
> `target_commitish` string
> Specifies the commitish value that determines where the Git tag is created from. Can be any branch or commit SHA. Unused if the Git tag already exists. Default: the repository's default branch.
> ```
> 
> AFAIU, this means that the create release REST API creates a tag. Wdyt?

A: thanks for the confirmation.

adglkh (approved):
LGTM

morphis (approved):

PR #123, opened 2025-02-26
Respect verticalAlignment setting for video top offset

adglkh:
With the introduction of the verticalAlignment setting, the video's top 
offset should be adjusted accordingly. Otherwise, input event positions
may be misaligned.

Fixes https://bugs.launchpad.net/anbox-cloud/+bug/2093904

### DIFF
--- a/js/tests/unit/sdk.test.js
+++ b/js/tests/unit/sdk.test.js
@@ -467,3 +467,38 @@ test("do no stop streaming on non fatal errors", (done) => {
   expect(() => stream.disconnect()).not.toThrow();
   expect(stream._webrtcManager.stop).toHaveBeenCalledTimes(1);
 });
+
+test("player respects the vertical aligment settings", () => {
+  const verticalAlignments = ["top", "center", "bottom"];
+
+  for (const alignment of verticalAlignments) {
+    let opts = { ...sdkOptions };
+    opts.verticalAlignment = alignment;
+    const stream = new AnboxStream(opts);
+
+    const video = document.createElement("video");
+    video.id = stream._videoID;
+    video.__defineGetter__("videoWidth", () => 1000);
+    video.__defineGetter__("videoHeight", () => 500);
+
+    const container = document.getElementById(sdkOptions.targetElement);
+    container.__defineGetter__("clientWidth", () => 2000);
+    container.__defineGetter__("clientHeight", () => 2000);
+    container.appendChild(video);
+
+    stream._onResize();
+    let dimensions = stream._dimensions;
+    expect(dimensions.playerWidth).toEqual(2000);
+    expect(dimensions.playerHeight).toEqual(1000);
+
+    let expectedOffsetTop;
+    if (alignment === "top") {
+      expectedOffsetTop = 0;
+    } else if (alignment === "center") {
+      expectedOffsetTop = 500;
+    } else if (alignment === "bottom") {
+      expectedOffsetTop = 1000;
+    }
+    expect(dimensions.playerOffsetTop).toEqual(expectedOffsetTop);
+  }
+});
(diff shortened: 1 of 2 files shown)


js/tests/unit/sdk.test.js
```diff
@@ -467,3 +467,39 @@ test("do no stop streaming on non fatal errors", (done) => {
...
+
+    let expectedOffsetTop;
+    if (alignment === "top") {
+      expectedOffsetTop = 0;
+    } else if (alignment === "center") {
+      expectedOffsetTop =
+        (container.clientHeight - dimensions.playerHeight) / 2;
+    } else if (alignment === "bottom") {
+      expectedOffsetTop = container.clientHeight - dimensions.playerHeight;
+    }
```
lorumic, line 502: For the type of stream that you have created here (height is 2x width), the `expectedOffsetTop` will be 0 in all 3 cases. I think this is not very representative, we should swap width and height to simulate a stream that is wider than taller (i.e. a "landscape-oriented" stream).
In other words, with this kind of stream "shape" the `verticalAlignment` value doesn't really have an effect.

adglkh: As you can see, it is also a valid case when the stream window is in portrait mode. I don’t mind taking your suggestion, swapping the width and height, and setting expectedOffsetTop to an explicit values.

js/tests/unit/sdk.test.js
```diff
@@ -467,3 +467,38 @@ test("do no stop streaming on non fatal errors", (done) => {
...
+    expect(dimensions.playerWidth).toEqual(2000);
+    expect(dimensions.playerHeight).toEqual(1000);
+
+    let expectedOffsetTop;
+    if (alignment === "top") {
+      expectedOffsetTop = 0;
+    } else if (alignment === "center") {
+      expectedOffsetTop = 500;
+    } else if (alignment === "bottom") {
+      expectedOffsetTop = 1000;
```
lorumic, line 500: Why the hard-coded values here? Can't we keep the calculations as in the previous iteration?

adglkh: Using explicit values here in unit tests instead of calculations because the dynamic calculations introduce a risk of hidden dependencies or unintended changes affecting the test outcome. (The case you mentioned above has been the case, where the expectedOffsetTop will be 0 in all 3 cases in portrait mode). Explicit values also makes test case immediately clear what the expected result is and easier to understand. 
That's being said, the test uses the same logic as the implementation to compute expected results, it may not catch actual errors in the code.

lorumic: Makes sense, thanks!

lorumic (approved):
LGTM - we should complement this with visual e2e tests as well, but it can be out of the scope of this PR. I've filed [WD-19671](https://warthogs.atlassian.net/browse/WD-19671) for this.

js/tests/unit/sdk.test.js
lorumic: > it is also a valid case when the stream window is in portrait mode

Sure, but this is not the current case of the test, because the stream container has a landscape (width > height) shape. Thanks for the changes, resolving this.

adglkh:
> LGTM - we should complement this with visual e2e tests as well, but it can be out of the scope of this PR. I've filed [WD-19671](https://warthogs.atlassian.net/browse/WD-19671) for this.

A: Okay, thanks. One more question: I can't seem to find a way to influence the vertical alignment from the dashboard. Is this something we'd like to add, or am I missing something?

lorumic:
> I can't seem to find a way to influence the vertical alignment from the dashboard.

In the dashboard we just need to pass the `verticalAlignment` option when we initialise the `AnboxStream`, [here](https://github.com/canonical/ams/blob/main/ui/templates/js/pages/sessions/%5BsessionId%5D/SessionStream.tsx#L169).

> Is this something we'd like to add, or am I missing something?

We do. That's a request from our visual team. Vertical alignment should be changed to `top` instead of the default `center`.

PR #130, opened 2025-03-18
chore: use global TICSAUTHTOKEN

morphis:
## Done

There is now a secret present at the org level named TICSAUTHTOKEN so we no longer need to have our own.

## QA

Run TICS workflow once merged

## JIRA / Launchpad bug

https://warthogs.atlassian.net/browse/AC-3245

## Documentation

n/a


ajanon (approved):
LGTM

PR #141, opened 2025-04-30
fix(gh): add pull-requests write permission to branching workflow

jat-canonical:
## Done

* This change allows the branching workflow to use github token to open a pull request on the repository itself.


ajanon (approved):
LGTM

PR #146, opened 2025-05-23
Update disk size in e2e tests

ajanon:
## Done

Update the minimum disk size in e2e tests to 4GB, following a change in AMS.

## QA

Run e2e tests.

## JIRA / Launchpad bug

N/A

## Documentation

N/A


lorumic (approved):
LGTM, thanks for the fix!

PR #149, opened 2025-07-08
fix(WebRTCManager): assign a callback to WebRTCManager only when defined

lorumic:
## Done

- Assign a callback to WebRTCManager only when defined

## QA

- Instantiate an `AnboxStream` without providing, in the `options`, one of the WebRTCManager callbacks, e.g. `onStatsUpdated`
- Join the session stream
- Before this change, you would get this in the console:
```
anbox-stream-sdk.js:3325 Uncaught (in promise) TypeError: this._onStatsUpdated is not a function
    at anbox-stream-sdk.js:3325:14
```
https://github.com/canonical/anbox-streaming-sdk/blob/c9d6f5d0df2aa66cc2ec05df2d1deb905beb3cfb/js/anbox-stream-sdk.js#L3323-L3327
- After this change, no error should be shown in the console, because we would not make the assignment when the callback is nullish, hence falling back to the default empty function of the `AnboxWebRTCManager` class:
https://github.com/canonical/anbox-streaming-sdk/blob/c9d6f5d0df2aa66cc2ec05df2d1deb905beb3cfb/js/anbox-stream-sdk.js#L2342


js/anbox-stream-sdk.js
```diff
@@ -168,14 +168,22 @@ class AnboxStream {
...
-    this._webrtcManager.onCameraRequested(
-      this._options.callbacks.requestCameraAccess,
-    );
-    this._webrtcManager.onMicrophoneRequested(
-      this._options.callbacks.requestMicrophoneAccess,
-    );
+    if (this._options.callbacks.statsUpdated) {
+      this._webrtcManager.onStatsUpdated(this._options.callbacks.statsUpdated);
+    }
+    if (this._options.callbacks.messageReceived) {
```
adglkh, line 174: At this point, the callback functions should be properly initialized, since AnboxStream._fillDefaults() is invoked during the AnboxStream object's constructor.
https://github.com/canonical/anbox-streaming-sdk/blob/df2ff83c8d1272d691d8b539603cedd719215961/js/anbox-stream-sdk.js#L676-L683

It appears we need to add a default dummy function for the statsUpdated callback to handle cases where it's undefined.

lorumic: Ah, I see - so `statsUpdated` was simply missing in the `_fillDefaults` function! Thanks for pointing it out @adglkh - let me fix it there then.

lorumic: [Updated](https://github.com/canonical/anbox-streaming-sdk/compare/df2ff83c8d1272d691d8b539603cedd719215961..12eb118a1fa13548893ffd9c3ceeb2b93fb1afe3)

anusha-c18 (approved):
lgtm

PR #159, opened 2025-09-09
Replace softprops action with gh CLI call

ajanon:
## Done

Replace the softprops action with a direct call to the GitHub CLI tool. This should work in the same way as before, and decreases our reliance on external action.

## QA

Run Release workflow and check that it still works as intended. We need to test this before the next release.

## JIRA / Launchpad bug

N/A

## Documentation

N/A


morphis (approved):
Nice, LGTM

PR #162, opened 2025-09-28
Sensor support

adglkh:
Add `AnboxSensorManager` class to capture and process device sensor data
using the HTML5 `DeviceOrientationEvent` and `DeviceMotionEvent` APIs:

- https://developer.mozilla.org/en-US/docs/Web/API/DeviceOrientationEvent
- https://developer.mozilla.org/en-US/docs/Web/API/DeviceMotionEvent

These APIs cover three major sensors:

- Accelerometer
- Gyroscope
- Orientation

Sensor data from the client is sent to the Android container once the
WebRTC connection is established, allowing Android applications that
rely on sensor input to function correctly. When the WebRTC connection
is stopped, the AnboxSensorManager stops capturing sensor data.

Other sensors are currently not implemented, either due to HTML5 API
limitations or little interests:

- Magnetometer
- Temperature
- Proximity
- Light
- Pressure
- Humidity

## QA

1. Follow the tutorial to setup a stream client
    https://documentation.ubuntu.com/anbox-cloud/tutorial/stream-client/
2. Construct anbox stream object with sensors enabled
```
const stream = new AnboxStream({
  connector: connector,
  targetElement: "anbox-stream",
  enableStats: false,
  devices: {
    sensor: {
      enableGyroscope: true,
      enableAccelerometer: true,
      enableOrientation: true,
    }
   }
});
```
3. Launch a streamable client and install a [sensor test](https://play.google.com/store/apps/details?id=ru.andr7e.sensortest&hl=en_GB) app on it
     amc launch --devmode --enable-streaming

4. Once the client starts streaming, move or tilt the device in different directions and verify that the values of the following sensors change accordingly:
- Accelerometer
- Gyroscope
- Orientation

## JIRA / Launchpad bug

Fixes AC-3756

## Documentation

A: yes, in 1.28.0 release.


js/tests/unit/sensor-events.test.js
```diff
@@ -0,0 +1,391 @@
...
+  sdkOptions.connector = { connect() {}, disconnect() {} };
+  sdkOptions.targetElement = "foobar";
+  sdkOptions.controls = { emulateTouch: false };
+  global.navigator.__defineGetter__("maxTouchPoints", () => 5);
+  global.navigator.__defineGetter__(
+    "userAgent",
+    () =>
+      "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/98.0.4758.109 Safari/537.36",
+  );
+});
```
lorumic, line 93: Ideally, since this is all reusable code that is currently duplicated across tests, we should extract into a module and import it wherever it is needed.

js/tests/unit/sensor-events.test.js
```diff
@@ -0,0 +1,391 @@
...
+      updateInterval: 5,
+      enableOrientation: true,
+    },
+  };
+  expect(() => new AnboxStream(sdkOptions)).toThrow(
+    "update interval 5ms is less than the minimum of 10ms",
+  );
+});
+
+test("should not send sensor data to Anbox container if all sensors are disabled", () => {
```
lorumic, line 107: Maybe we should also add a test for the defaults, i.e. when the `sensor` option is not provided at all?

js/tests/unit/sensor-events.test.js
```diff
@@ -0,0 +1,391 @@
...
+  expect(snapshots[0].type).toEqual("sensor:event");
+  expect(snapshots[0].data).toEqual({
+    sensor: "acceleration",
+    x: 1.0,
+    y: 2.5,
+    z: 3.0,
+  });
+  // gyroscope or orientation data should be availabe in the payload as both are disabled.
+  expect(snapshots[0].type).not.toHaveProperty("gyroscope");
+  expect(snapshots[0].type).not.toHaveProperty("orientation");
```
lorumic, line 145: I don't understand these checks. `snapshots[0].type` is a string (see the check above at line 136), so how can it have properties? Maybe you meant `snapshots[0].data`, but then again that is not ever expected to contain these as properties in the payload, since the format is `sensor: <sensor-name>, ...`, so none of `acceleration`, `gyroscope` or `orientation` will ever appear in the payload as a property.

js/tests/unit/sensor-events.test.js
```diff
@@ -0,0 +1,391 @@
...
+      enableAccelerometer: true,
+    },
+  };
+  let { stream, snapshots } = setupStream(sdkOptions);
+  expect(stream._sensorManager).not.toBeNull();
+  stream._webrtcReady(null, null);
+
+  window.dispatchEvent(
+    new DeviceMotionEvent("devicemotion", {
+      accelerationIncludingGravity: { x: 1.0, y: 2.5, z: 3.0 },
```
lorumic, line 131: Can we simulate a case where `acceleration` is used instead of `accelerationIncludingGravity`?

js/tests/unit/sensor-events.test.js
```diff
@@ -0,0 +1,391 @@
...
+    new DeviceMotionEvent("devicemotion", {
+      rotationRate: {
+        alpha: 10.0,
+        beta: 10.0,
+        gamma: 10.0,
+      },
+    }),
+  );
+
+  ensure_signal_update();
```
lorumic, line 256: After this, maybe we can advance 1 second to check that the new data was sent eventually?

js/tests/unit/sensor-events.test.js
adglkh: +1

js/tests/unit/sensor-events.test.js
adglkh: I could add one.

js/anbox-stream-sdk.js
```diff
@@ -207,6 +222,22 @@ class AnboxStream {
...
+    this._sensorManager = null;
+    const sensorsEnabled =
+      this._options.devices.sensor.enabled ||
+      this._options.devices.sensor.enableOrientation ||
+      this._options.devices.sensor.enableAccelerometer ||
+      this._options.devices.sensor.enableGyroscope;
+    if (sensorsEnabled) {
+      this._sensorManager = new AnboxSensorManager({
+        webrtcManager: this._webrtcManager,
+        updateInterval: options.devices.sensor.updateInterval,
```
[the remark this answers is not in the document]
adglkh: Thanks for mentioning here.

.gitignore
```diff
@@ -1,3 +1,4 @@
 .npm
 node_modules
-.idea
\ No newline at end of file
+.idea
+js/tests/unit/anbox-stream-sdk.js
```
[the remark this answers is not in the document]
adglkh: Nope, we copied it 
https://github.com/adglkh/anbox-streaming-sdk/blob/511370f953e270964961ff58b9188985c7cc8ee0/js/tests/unit/run-tests.sh#L42
before running the unit test

js/tests/unit/sensor-events.test.js
lorumic: Ok, let's do that.

js/anbox-stream-sdk.js
```diff
@@ -4423,6 +4476,227 @@ class AnboxVhalManager {
...
+   */
+  constructor(options = {}) {
+    let updateInterval =
+      options.updateInterval ?? DEFAULT_SENSOR_DATA_UPDATE_INTERVAL;
+    if (typeof updateInterval !== "number") {
+      throw newError(
+        `sensor update interval is invalid`,
+        ANBOX_STREAM_SDK_ERROR_INVALID_ARGUMENT,
+      );
+    }
```
lorumic, line 4491: Nice, since we're here, can we also add a unit test that covers this specific edge case of invalid (non-numeric) value passed?

js/anbox-stream-sdk.js
```diff
@@ -4423,6 +4476,227 @@ class AnboxVhalManager {
...
+        error,
+      );
+    }
+  }
+
+  _onDeviceOrientation(event) {
+    if (!this._options.enableOrientation) return;
+
+    if (!this.sensorData.orientation) {
+      this.sensorData.orientation = { roll: null, pitch: null, azimuth: null };
```
lorumic, line 4599: With these default values set to `null`, we should make sure to adjust the `fuzzyCompare` function accordingly, or we would miss an update when we have `null` and we get a sensor value of `0`, because `null` coalesces with `0`. We should make sure to add a check `if (n1 === null || n2 === null) return false;` there.

js/tests/unit/sensor-events.test.js
```diff
@@ -0,0 +1,444 @@
...
+      updateInterval: 5,
+      enableOrientation: true,
+    },
+  };
+  expect(() => new AnboxStream(sdkOptions)).toThrow(
+    "update interval 5ms is less than the minimum of 10ms",
+  );
+});
+
+test("should not be construct sensor manager if it is not specified in the options.", () => {
```
lorumic, line 102:
```suggestion
test("should not init the sensor manager if it is not specified in the options.", () => {
```

js/tests/unit/sensor-events.test.js
```diff
@@ -0,0 +1,444 @@
...
+  expect(snapshots[0].type).toEqual("sensor:event");
+  expect(snapshots[0].data).toEqual({
+    sensor: "acceleration",
+    x: 1.0,
+    y: 2.5,
+    z: 3.0,
+  });
+  // gyroscope or orientation data should not be availabe in the payload as both are disabled.
+  expect(snapshots[0].data.sensor).not.toBe("gyroscope");
+  expect(snapshots[0].data.sensor).not.toBe("orientation");
```
lorumic, line 146: I think this is redundant - we already check the same thing above with:
```
  expect(snapshots.length).toEqual(1);
  expect(snapshots[0].type).toEqual("sensor:event");
  expect(snapshots[0].data).toEqual({
    sensor: "acceleration",
    x: 1.0,
    y: 2.5,
    z: 3.0,
  });
```
```suggestion
```
And the same should be applied to the other tests as well.

js/tests/unit/sensor-events.test.js
```diff
@@ -0,0 +1,444 @@
...
+      },
+    }),
+  );
+
+  ensure_signal_update();
+
+  // Advance 1 second
+  fakeTime += 1000;
+
+  // Capture data from accleration sensor
```
lorumic, line 295:
```suggestion
  // Capture data from acceleration sensor
```

lorumic (changes requested):
QA LGTM, a few other small comments for the code and then we're good to go.

js/anbox-stream-sdk.js
adglkh: hmm, I thought I added. 
Thanks for spotting this.

.gitignore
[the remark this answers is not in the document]
adglkh: Yes, I think it's harmless if anbox-stream-sdk.js remains in that folder as a leftover. We don’t need to adjust the git command to remove and re-stage it. What do you think?

ajanon (approved):
LGTM

PR #209, opened 2026-04-17
Add the missing back key

adglkh:
## Done

This enables the simulation of Back key events sent from the WebRTC
client and being forwarded correctly to Android system over the control
data channel.

NOTE: Anbox WebRTC platform already supports the Back key on the backend.
This change implements the corresponding client-side support to complete
the e2e integration.

## QA

1. Open up an application E.g. settings
2. Fire the following key event from a stream client 

```
stream = new AnboxStream({...})
stream.sendInputKey("Back");
```
3. Ensure that the active application is pushed to the background

## JIRA / Launchpad bug

Fixes #

## Documentation

A:no, it doesn't

### DIFF
--- a/js/anbox-stream-sdk.js
+++ b/js/anbox-stream-sdk.js
@@ -2161,6 +2161,7 @@ const _keyScancodes = {
   Volumedown: 89,
   Volumeup: 90,
   Power: 91,
+  Back: 92,
 };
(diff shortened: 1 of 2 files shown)


anusha-c18 (approved):
lgtm :D

PR #213, opened 2026-05-28
fix(sdk): forward Enter key to Android container to allow IME dismissal

adglkh:
## Done

Some games and apps rely sendKeyEvent(KEYCODE_ENTER) or on
performEditorAction()(IME_ACTION_DONE, IME_ACTION_GO, etc.) or to
dismiss client side virtual keyboard. These paths were not forwarded
to the Android container, so AnboxIME never received a signal to hide,
causing the keyboard to remain visible after the user pressed Enter.
    
To resolve this, a new onEnterKeyEvent() callback has been added to
TextChangedListener and implemented in InputConnection. AnboxWebView
leverages this callback to invoke the newly exposed sendIMEKeyCode()
JS API, bridging the key event directly to the stream.

## QA

1. During gameplay (e.g., entering the player's name in Homescapes), the client-side 
   virtual keyboard now properly dismisses when the user hits "Enter" key.
2. Verified that this change doesn't introduce any regressions to text input in other games.

## JIRA / Launchpad bug

Fixes https://warthogs.atlassian.net/browse/AC-4526

## Documentation

A: yes, in the 1.30.0 release announcement.

### DIFF
--- a/android/anbox_streaming_sdk/src/main/java/com/canonical/anbox/streaming_sdk/AnboxWebView.java
+++ b/android/anbox_streaming_sdk/src/main/java/com/canonical/anbox/streaming_sdk/AnboxWebView.java
@@ -94,6 +94,14 @@ public class AnboxWebView extends WebView implements
          *        ratio on the client side.
          */
         void onVirtualKeyboardStateChanged(boolean isShow, double heightRatio);
+
+        /**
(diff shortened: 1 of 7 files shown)


lorumic (approved):
Code LGTM, but unfortunately I do not have the resources to test it/do the QA.

PR #215, opened 2026-06-03
Log full ICE candidate

adglkh:
## Done

Include candidate string in the log along with sdpMid and sdpMLineIndex, 
so that the candidate type, address, and associated media section are visible
when `options.experimental.debug` is set to true.

## QA

Ensure the ICE candidate is printed out in the console when
`options.experimental.debug` option is set to true.

## JIRA / Launchpad bug

Fixes #

## Documentation

A: no, it doesn't


lorumic (approved):
LGTM 👍

PR #216, opened 2026-06-11
fix(vhal-manager): increase default timeout from 1s to 5s

lorumic:
## Done

- Increased default timeout for the VhalManager from 1s to 5s
- The 1s timeout was insufficient for the new euphotic AAOS images


adglkh (approved):
LGTM

tarek-hanna (approved):

PR #219, opened 2026-07-15
Multi displays support

adglkh:
## Done

This is the initial multi-display support in the JS SDK, which enabling the
WebRTC manager to handle multiple simultaneous video tracks, touch 
tracking and input event routing per-display.

The targetElement (video container) is now optional, preserved for backward 
compatibility in legacy single-display mode. 

Two new callbacks
- videoTrackAdded
- videoTrackRemoved

are added in the JS SDK to support dynamical display attachment(multi-display mode)
via attachDisplay() API.

## QA

1. Enable newly added tests cases pass
2. For the real e2e testing, please try out the [webrtc_multiple_display](https://github.com/adglkh/ams/tree/webrtc_multiple_display), which has this JS SDK mirrored in the dashboard UI
    - Launch an instance multi displays
          $ amc launch resolute:android16-cf:amd64 --enable-streaming --devmode -c 4 -m 8GB --name test-multi-display \\
             --display=width=1280,height=720,fps=60 \\
             --display=width=480,height=640,dpi=120 \\
      Once the instance is up and running, ensure each display are present on the UI properly
    - In the meanwhile, present different apps in different displays within the instance
        $ amc exec <instance_id> --  adb shell am start-activity -W -n com.android.dialer/.main.impl.MainActivity --display 2
        $ amc exec <instance_id> --  adb shell am start-activity -W -n com.android.gallery3d/com.android.gallery3d.app.Gallery --display 3 

## JIRA / Launchpad bug

Fixes https://warthogs.atlassian.net/browse/AC-4542

## Documentation

A: yes, in the 1.31.0 release announcement.

### DIFF
--- a/js/anbox-stream-sdk.js
+++ b/js/anbox-stream-sdk.js
@@ -158,10 +158,33 @@ class AnboxStream {
       this._detectUnsupportedBrowser();
 
     this._id = Math.random().toString(36).substr(2, 9);
-    this._containerID = options.targetElement;
+
+    // _containerID is the primary display container. It's reserved for
+    // in single display.
+    this._containerID = options.targetElement || null;
+    // _containerIDs is an array of video container IDs used to attach input listeners
+    // to multiple video tracks.
+    this._containerIDs = this._containerID ? {0: this._containerID} : {};
+
     this._videoID = "anbox-stream-video-" + this._id;
     this._audioID = "anbox-stream-audio-" + this._id;
     this._canvasID = "anbox-stream-canvas-" + this._id;
+    this._multiDisplayActive = false;
+
+    this._pendingVideoTracks = {};
+    this._pendingReadyCount = 0;
+
+    this._displayStates = this._containerID
+      ? {
+          0: {
+            dimensions: null,
+            activeTouchPointers: [],
+            pointerIdsMapper: {},
+            primaryTouchId: 0,
+            pointersOutofBounds: {},
+          },
+        }
+      : {};
 
     // WebRTC
     this._webrtcManager = new AnboxWebRTCManager({
@@ -248,16 +271,45 @@ class AnboxStream {
 
     // Control options
     this._modifierState = 0;
-    this._dimensions = null;
     this._gamepadManager = null;
     this._streamCanvas = null;
 
     this._originalOrientation = null;
     this._currentRotationDegrees = 0;
-    this._primaryTouchId = 0;
-    this._pointersOutofBounds = {};
-    this._activeTouchPointers = [];
-    this._pointerIdsMapper = {};
+    // Legacy aliases forward to primary display state for any code that was not
+    // yet updated to use the per-display state bag.
+    Object.defineProperty(this, "_dimensions", {
+      get: () => this._displayStates[0].dimensions,
+      set: (v) => {
+        this._displayStates[0].dimensions = v;
+      },
+    });
+    Object.defineProperty(this, "_primaryTouchId", {
+      get: () => this._displayStates[0].primaryTouchId,
+      set: (v) => {
+        this._displayStates[0].primaryTouchId = v;
+      },
+    });
+    Object.defineProperty(this, "_pointersOutofBounds", {
+      get: () => this._displayStates[0].pointersOutofBounds,
+      set: (v) => {
+        this._displayStates[0].pointersOutofBounds = v;
+      },
+    });
+    Object.defineProperty(this, "_activeTouchPointers", {
+      get: () => this._displayStates[0].activeTouchPointers,
+      set: (v) => {
+        this._displayStates[0].activeTouchPointers = v;
+      },
+    });
+    Object.defineProperty(this, "_pointerIdsMapper", {
+      get: () => this._displayStates[0].pointerIdsMapper,
+      set: (v) => {
+        this._displayStates[0].pointerIdsMapper = v;
+      },
+    });
+
+    this._displayEventListeners = [];
 
     this.controls = {
       touch: {
@@ -1010,17 +1062,38 @@ class AnboxStream {
     window.addEventListener("resize", this._onResize);
 
     if (this._options.controls.mouse) {
-      const container = document.getElementById(this._containerID);
-      if (container) {
-        for (const controlName in this.controls.touch)
-          container.addEventListener(
-            controlName,
-            this.controls.touch[controlName],
-          );
-      }
+      // Register primary display input listeners on its container.
+      const container = document.getElementById(this._containerIDs[0]);
+      if (container) this._registerInputHandlers(0, container);
     }
 
     this.captureKeyboard();
+  }
+
+  _registerInputHandlers(displayId, container) {
+    if (!this._options.controls.mouse) return;
+
+    const handlers = {
+      pointermove: (e) => this._onPointerEvent(e, displayId),
+      pointerdown: (e) => this._onPointerEvent(e, displayId),
+      pointerup: (e) => this._onPointerEvent(e, displayId),
+      pointercancel: (e) => this._onPointerEvent(e, displayId),
+      mousewheel: (e) => this._onMouseWheel(e, displayId),
+    };
+
+    for (const [name, fn] of Object.entries(handlers))
+      container.addEventListener(name, fn);
+
+    this._displayEventListeners[displayId] = { container, handlers };
+  }
+
+  _unregisterInputHandlers(displayId) {
+    const entry = this._displayEventListeners[displayId];
+    if (!entry) return;
+    const { container, handlers } = entry;
+    for (const [name, fn] of Object.entries(handlers))
+      container.removeEventListener(name, fn);
+    delete this._displayEventListeners[displayId];
   }
 
   /**
@@ -1148,17 +1221,9 @@ class AnboxStream {
   _unregisterControls() {
     window.removeEventListener("resize", this._onResize);
 
-    // Removing the video container should automatically remove all event listeners
-    // but this is dependant on the garbage collector, so we manually do it if we can
     if (this._options.controls.mouse) {
-      const container = document.getElementById(this._containerID);
-      if (container) {
-        for (const controlName in this.controls.touch)
-          container.removeEventListener(
-            controlName,
-            this.controls.touch[controlName],
-          );
-      }
+      for (const i of Object.keys(this._containerIDs).map(Number))
+        this._unregisterInputHandlers(i);
     }
 
     this.releaseKeyboard();
@@ -1299,7 +1364,28 @@ class AnboxStream {
     if (video === null || container === null) return;
     if (video.videoWidth === 0 || video.videoHeight === 0) return;
 
-    // We calculate the distance to the closest window border while keeping aspect ratio intact.
+    // In multi-display mode, recompute all attached displays so that any
+    // layout changes (new display added, window resize) are reflected.
+    if (this._multiDisplayActive) {
+      this._computeMultiDisplayDimensions(video, container, 0);
+      const extraIds = Object.keys(this._containerIDs).map(Number).filter(i => i > 0);
+      for (const i of extraIds) {
+        const secId = `${this._videoID}-display-${i}`;
+        const secVideo = document.getElementById(secId);
+        const secContainer = document.getElementById(this._containerIDs[i]);
+        if (
+          secVideo &&
+          secContainer &&
+          secVideo.videoWidth > 0 &&
+          secVideo.videoHeight > 0
+        ) {
+          this._computeMultiDisplayDimensions(secVideo, secContainer, i);
+        }
+      }
+      return;
+    }
+
+    // Single-display mode
     let videoHeight = video.videoHeight;
     let videoWidth = video.videoWidth;
 
@@ -1400,7 +1486,7 @@ class AnboxStream {
     // The visual offset is always derived from the same formula, no matter the orientation.
     const offsetLeft = Math.round(container.clientWidth / 2 - playerWidth / 2);
 
-    this._dimensions = {
+    this._displayStates[0].dimensions = {
       videoHeight: videoHeight,
       videoWidth: videoWidth,
       scalePercentage: resizePercentage,
@@ -1408,6 +1494,52 @@ class AnboxStream {
       playerWidth: playerWidth,
       playerOffsetLeft: offsetLeft,
       playerOffsetTop: offsetTop,
+    };
+  }
+
+  // Compute dimensions for a single display's video element inside its container
+  // for multi-display mode.
+  _computeMultiDisplayDimensions(video, container, displayId) {
+    const cRect = container.getBoundingClientRect();
+    const cellWidth = cRect.width;
+    const cellHeight = cRect.height;
+
+    const videoWidth = video.videoWidth;
+    const videoHeight = video.videoHeight;
+
+    // Compute the largest size that fits the video aspect ratio inside the cell.
+    const scale = Math.min(cellWidth / videoWidth, cellHeight / videoHeight);
+    const renderedWidth = Math.round(videoWidth * scale);
+    const renderedHeight = Math.round(videoHeight * scale);
+    const offsetLeft = Math.round((cellWidth - renderedWidth) / 2);
+    const offsetTop = Math.round((cellHeight - renderedHeight) / 2);
+
+    // Size and position the video element to exactly match the rendered content.
+    video.style.width = renderedWidth + "px";
+    video.style.height = renderedHeight + "px";
+    video.style.left = offsetLeft + "px";
+    video.style.top = offsetTop + "px";
+
+    if (!(displayId in this._displayStates)) {
+      this._displayStates[displayId] = {
+        dimensions: null,
+        activeTouchPointers: [],
+        pointerIdsMapper: {},
+        primaryTouchId: 0,
+        pointersOutofBounds: {},
+      };
+    }
+
+    // The playerOffset is 0 because the video element is already sized to the
+    // rendered content.
+    this._displayStates[displayId].dimensions = {
+      videoHeight: videoHeight,
+      videoWidth: videoWidth,
+      scalePercentage: scale,
+      playerHeight: renderedHeight,
+      playerWidth: renderedWidth,
+      playerOffsetLeft: 0,
+      playerOffsetTop: 0,
     };
   }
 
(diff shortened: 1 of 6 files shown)


js/tests/unit/multi-display.test.js
```diff
@@ -0,0 +1,555 @@
...
+    stream.attachDisplay(0, "main-container");
+
+    const video = document.getElementById(stream._videoID);
+    expect(video).not.toBeNull();
+    expect(video.srcObject).toBe(fakeStream);
+    expect(stream._containerIDs[0]).toBe("main-container");
+    expect(stream._pendingVideoTracks[0]).toBeUndefined();
+  });
+
+  test("attachDisplay(1, containerId) injects secondary video and registers input handlers", () => {
```
lorumic, line 282: This doesn't seem to test whether the input handlers are registered, or am I missing something?

```suggestion
  test("attachDisplay(1, containerId) injects secondary video", () => {
```

js/tests/unit/multi-display.test.js
```diff
@@ -0,0 +1,555 @@
...
+      }),
+    );
+
+    expect(mockSend).toHaveBeenCalledWith(
+      "input::mouse-button",
+      expect.objectContaining({ display_id: 1 }),
+    );
+  });
+
+  test("events on display 0 container include display_id: 0", () => {
```
lorumic: Do we need this test? Seems redundant as it duplicates the test above entirely. Also code duplication should be avoided - if we really need 2 tests, we should have a reusable function which accepts what changes in the 2 tests as parameters, and reuse it in the 2 occurrences.

js/tests/unit/multi-display.test.js
```diff
@@ -0,0 +1,555 @@
...
+
+  test("resets _containerIDs and _displayStates to initial values", () => {
+    const stream = makeStreamWithTarget("main-container");
+    stream._containerIDs = { 0: "main-container", 1: "cell-1" };
+
+    stream._removeMedia();
+
+    expect(stream._containerIDs).toEqual({ 0: "main-container" });
+    // Should have one display state entry (for display 0).
+    expect(0 in stream._displayStates).toBe(true);
```
lorumic, line 492: Should we also verify the absence of the other display state entry?

```suggestion
    expect(0 in stream._displayStates).toBe(true);
    expect(1 in stream._displayStates).toBe(false);
```

js/tests/unit/multi-display.test.js
```diff
@@ -0,0 +1,555 @@
...
+    const stream = makeStream({
+      targetElement: "main-container",
+      callbacks: { videoTrackRemoved: (i) => removed.push(i) },
+    });
+
+    stream._containerIDs = { 0: "main-container", 1: "cell-1", 2: "cell-2" };
+    stream._removeMedia();
+
+    expect(removed).toEqual([2, 1, 0]);
+  });
```
lorumic: This fully duplicates the test at line 436. Can we remove it?

```suggestion
```

js/anbox-stream-sdk.js
```diff
@@ -86,7 +86,12 @@ class AnboxStream {
    * displays its video & audio feed in an HTML5 player
    * @param options: {object}
    * @param options.connector {object} WebRTC Stream connector.
-   * @param options.targetElement {string} ID of the DOM element to attach the video to.
+   * @param [options.targetElement] {string} ID of the DOM element to attach the primary (display 0)
+   *   video to. When provided the SDK operates in legacy single-display mode: display 0 is
+   *   injected automatically, and the `videoTrackAdded` callback fires for every display that arrives
+   *   (including display 0). When omitted the SDK operates in fully dynamic mode: the caller MUST
+   *   call `attachDisplay(displayId, containerId)` for every display, including display 0, in
+   *   response to the `videoTrackAdded` callback. Array values are not supported.
```
lorumic: We might not need this specification. The type `{string}` should remove any doubt.

```suggestion
   *   response to the `videoTrackAdded` callback.
```

js/anbox-stream-sdk.js
```diff
@@ -316,6 +347,130 @@ class AnboxStream {
...
+        { once: true },
+      );
+
+      // Recompute when the container cell resizes.
+      if (typeof ResizeObserver !== "undefined") {
+        const ro = new ResizeObserver(() => {
+          if (video.videoWidth > 0 && video.videoHeight > 0) computeDims();
+        });
+        ro.observe(container);
+      }
```
lorumic: Can we try to reuse the code from the if block above (lines 417-437) instead of duplicating it?

js/anbox-stream-sdk.js
```diff
@@ -809,24 +970,25 @@ class AnboxStream {
...
 
-    if (this._nullOrUndef(options.targetElement)) {
-      throw newError(
-        "missing targetElement parameter",
-        ANBOX_STREAM_SDK_ERROR_INVALID_ARGUMENT,
-      );
+    if (!this._nullOrUndef(options.targetElement)) {
+      if (typeof options.targetElement !== "string") {
+        throw newError(
+          "targetElement must be a string ID (array values are not supported)",
```
lorumic: Again, I think we don't need to specify explicitly that array values are not supported. The user might have passed any non-string type, and would be puzzled to read "array values are not supported" in such case.

```suggestion
          "targetElement must be a string ID",
```

js/anbox-stream-sdk.js
```diff
@@ -809,24 +970,25 @@ class AnboxStream {
...
+          ANBOX_STREAM_SDK_ERROR_INVALID_ARGUMENT,
+        );
+      }
+      const container = document.getElementById(options.targetElement);
+      if (container === null) {
+        throw newError(
+          `target element "${options.targetElement}" does not exist`,
+          ANBOX_STREAM_SDK_ERROR_INVALID_ARGUMENT,
+        );
+      } else if (container.clientWidth == 0 || container.clientHeight == 0)
```
lorumic: Nitpick - can we use strict equality comparisons consistently?

```suggestion
      } else if (container.clientWidth === 0 || container.clientHeight === 0)
```

js/anbox-stream-sdk.js
```diff
@@ -970,16 +1176,134 @@ class AnboxStream {
...
+    this._multiDisplayActive = true;
+
+    // Wrap the primary video in a dedicated cell div so it becomes one
+    // equal grid item alongside the additional display cells.
+    const cell = document.createElement("div");
+    cell.id = `${this._videoID}-cell-0`;
+    cell.className = "anbox-stream-cell";
+    cell.style.position = "relative";
+    cell.style.overflow = "hidden";
+    cell.style.background = "#000";
```
lorumic: Do we really need to enforce a black background for the cell? Shouldn't this be a cosmetic user's choice?

js/anbox-stream-sdk.js
```diff
@@ -1473,74 +1883,134 @@ class AnboxStream {
...
+        playerOffsetLeft: contentOffsetX,
+        playerOffsetTop: contentOffsetY,
+      };
+    } else {
+      state.dimensions.scalePercentage = scale;
+      state.dimensions.playerWidth = renderedWidth;
+      state.dimensions.playerHeight = renderedHeight;
+      state.dimensions.playerOffsetLeft = contentOffsetX;
+      state.dimensions.playerOffsetTop = contentOffsetY;
+    }
```
lorumic, line 1960: We can reuse some code here:

```suggestion
    const newValues = {
      scalePercentage: scale,
      playerWidth: renderedWidth,
      playerHeight: renderedHeight,
      playerOffsetLeft: contentOffsetX,
      playerOffsetTop: contentOffsetY,
    };
    if (!state.dimensions) {
      state.dimensions = {
        videoWidth: video.videoWidth,
        videoHeight: video.videoHeight,
        ...newValues,
      };
    } else {
      state.dimensions = { ...state.dimensions, ...newValues };
    }
```

js/anbox-stream-sdk.js
```diff
@@ -1473,74 +1883,134 @@ class AnboxStream {
...
    * Returns true if a pointer event (move or click) was emitted outside the video
-   * boundaries
+   * boundaries of the given display.
+   * @param {object} event
+   * @param {number} displayId
    * @returns {boolean}
    * @private
    */
-  _isPointerEventOutOfBounds(event) {
+  _isPointerEventOutOfBounds(event, displayId) {
```
lorumic, line 1874: I see that many methods are adding the `displayId` parameter now, but not all have a default value for it. Should we make it consistent and always default to 0 for full backward compatibility?

```suggestion
  _isPointerEventOutOfBounds(event, displayId = 0) {
```

js/anbox-stream-sdk.js
```diff
@@ -1473,74 +1883,134 @@ class AnboxStream {
...
   /**
-   * onPointerEvent is called when a mouse or touch input is fired
-   * @param pointerEvent
+   * onPointerEvent is called when a mouse or touch input is fired.
+   * @param pointerEvent {PointerEvent}
+   * @param {number} [displayId=0] Index of the display that received the event.
    * @private
    */
-  _onPointerEvent(pointerEvent) {
+  _onPointerEvent(pointerEvent, displayId = 0) {
```
lorumic, line 1998: Here, unlike the other methods above, we have a default value for `displayId` :)

js/anbox-stream-sdk.js
```diff
@@ -1551,6 +2021,9 @@ class AnboxStream {
     )
       return;
 
+    const state = this._displayStates[displayId];
```
lorumic: Maybe, just to be a bit clearer and improve code readability some tens of lines below:

```suggestion
    const displayState = this._displayStates[displayId];
```

js/anbox-stream-sdk.js
```diff
@@ -3284,16 +3788,40 @@ class AnboxWebRTCManager {
...
+      // MediaStream object for every track. A per-track MediaStream guarantees
+      // each video element renders its own display independently.
+      const singleTrackStream = new MediaStream([event.track]);
+      this._videoStreams[displayId] = singleTrackStream;
+
+      if (displayId === 0) {
+        this._videoStream = singleTrackStream;
+        if (event.streams[0]) event.streams[0].onremovetrack = this._onClose;
+      } else {
+        // External displays and notify AnboxStream to wire a new video element.
```
lorumic: I don't understand what this comment means (especially the first part "External displays"?) - can we make it a bit clearer please?

lorumic (commented):
Thank you @adglkh, this is really a monumental work. I tested its integration with the dashboard and everything worked really well, so I'm just leaving some minor comments, nitpicks, and ideas to optimise the code and avoid duplication. Excellent job overall! 💯

js/anbox-stream-sdk.js
```diff
@@ -3284,16 +3788,40 @@ class AnboxWebRTCManager {
...
-      this._videoStream.onremovetrack = this._onClose;
+      // The server assigns each video track the label "video_N" (where N is the
+      // display id) when adding video transceiver. On client side, parsing it
+      // here from MediaStreamTrack.id gives us the correct display id.
+      let displayId;
+      const m = event.track.id.match(/^video_(\d+)$/);
+      if (m) {
+        displayId = parseInt(m[1], 10);
+      } else {
+        console.error(`failed to paser display id: ${event.track.id}`);
```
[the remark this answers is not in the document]
adglkh: Thanks, fixed that!
When checking this piece of code, I found a regression, which the older images use the legacy track name "video" and we need to fix this.

js/anbox-stream-sdk.js
```diff
@@ -3284,16 +3788,40 @@ class AnboxWebRTCManager {
...
+      // the same stream_id on the server side so event.streams[0] is the same
+      // MediaStream object for every track. A per-track MediaStream guarantees
+      // each video element renders its own display independently.
+      const singleTrackStream = new MediaStream([event.track]);
+      this._videoStreams[displayId] = singleTrackStream;
+
+      if (displayId === 0) {
+        this._videoStream = singleTrackStream;
+        if (event.streams[0]) event.streams[0].onremovetrack = this._onClose;
+      } else {
```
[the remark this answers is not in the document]
adglkh: +1, for the primary display, when event.track.onended is fired, we probably need to trigger _onClose while for the external displays, we need to trigger videoTrackRemoved(displayId) instead .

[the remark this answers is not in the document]
adglkh: Fixed it

js/anbox-stream-sdk.js
```diff
@@ -970,16 +1176,134 @@ class AnboxStream {
...
+      primaryVideo.style.maxHeight = "";
+      cell.appendChild(primaryVideo);
     }
+    container.insertBefore(cell, container.firstChild);
+
+    // Move the display 0 input zone from the outer container to cell-0 so
+    // pointer events are scoped to the actual video area.
+    this._unregisterDisplayControls(0);
+    this._containerIDs[0] = cell.id;
+    this._registerInputHandlers(0, cell);
```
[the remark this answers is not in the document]
adglkh: +1, thanks for pointing this out.
Fixed it.

js/anbox-stream-sdk.js
```diff
@@ -1834,7 +2318,7 @@ class AnboxStream {
    * @private
    */
   _convertTouchInput(x, y) {
-    const dim = this._dimensions;
+    const dim = this._displayStates[0].dimensions;
     if (!dim) {
       throw newError("sdk is not ready yet", ANBOX_STREAM_SDK_ERROR_INTERNAL);
     }
```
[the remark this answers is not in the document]
adglkh: +1, Fixed that.
Thanks

js/anbox-stream-sdk.js
```diff
@@ -1473,74 +1883,134 @@ class AnboxStream {
...
    */
-  _translateLocalCoordsToRemoteCoords(event) {
+  _translateLocalCoordsToRemoteCoords(event, displayId) {
+    const dim = this._displayStates[displayId].dimensions;
+
     if (event.pointerType === "touch") {
       const pos = this._convertTouchInput(event.clientX, event.clientY);
       event.clientX = pos.x;
       event.clientY = pos.y;
     }
```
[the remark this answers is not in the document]
adglkh: See below.

js/anbox-stream-sdk.js
```diff
@@ -970,16 +1176,134 @@ class AnboxStream {
...
+      );
+      if (el) el.remove();
+    }
+
+    // Remove primary video and audio elements.
     const video = document.getElementById(this._videoID);
     const audio = document.getElementById(this._audioID);
-
     if (video) video.remove();
     if (audio) audio.remove();
```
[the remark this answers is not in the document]
adglkh: The underlyig issue was that the stale callback attached to the video container could be still fired after the session is torn down.
Make sense to me.
Fixed it

js/anbox-stream-sdk.js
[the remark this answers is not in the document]
adglkh: +1, please check the last commit.

js/anbox-stream-sdk.js
```diff
@@ -316,6 +347,130 @@ class AnboxStream {
...
+        { once: true },
+      );
+
+      // Recompute when the container cell resizes.
+      if (typeof ResizeObserver !== "undefined") {
+        const ro = new ResizeObserver(() => {
+          if (video.videoWidth > 0 && video.videoHeight > 0) computeDims();
+        });
+        ro.observe(container);
+      }
```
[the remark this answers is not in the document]
adglkh: Same below

js/anbox-stream-sdk.js
```diff
@@ -809,24 +970,25 @@ class AnboxStream {
...
-        "missing targetElement parameter",
-        ANBOX_STREAM_SDK_ERROR_INVALID_ARGUMENT,
-      );
+    if (!this._nullOrUndef(options.targetElement)) {
+      if (typeof options.targetElement !== "string") {
+        throw newError(
+          "targetElement must be a string ID (array values are not supported)",
+          ANBOX_STREAM_SDK_ERROR_INVALID_ARGUMENT,
+        );
+      }
```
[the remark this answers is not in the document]
adglkh: Make sense to me.
Thanks

js/tests/unit/multi-display.test.js
adglkh: Oh, we need to improve the test case to cover that.

adglkh: Done

js/tests/unit/multi-display.test.js
adglkh: Hmm, they're not redundant. Both of them covers the scenario where input events are routed to different displays (primary vs additional). But your point is valid that we should avoid code repetition.
Let me combine both

js/tests/unit/multi-display.test.js
adglkh: +1, make sense to me.

js/tests/unit/multi-display.test.js
adglkh: Indeed, thanks for pointing this out.

js/anbox-stream-sdk.js
adglkh: +1, yes,this is a leftover from when we agreed to keep `targetElement` for backward compatibility only.
Just dropped that.

js/anbox-stream-sdk.js
adglkh: +1, Done.
Thanks for pointing this out

js/anbox-stream-sdk.js
adglkh: Dropped that due to the leftover after the 1st iteration.

js/anbox-stream-sdk.js
adglkh: +1, Fixed that.

js/anbox-stream-sdk.js
adglkh: Agreed. That shouldn't be hardcoded and it's better to let users customize it.

js/anbox-stream-sdk.js
adglkh: Thanks for the suggestion!

js/anbox-stream-sdk.js
adglkh: Sure. fixed that.

js/anbox-stream-sdk.js
adglkh: Okay, this is about `Notify AnboxStream to wire a new video element for a display` as it internally 
AnboxWebRTCManager::_onRtcTrack when new video track is available 
|
AnboxWebRTCManager::_onExtraVideoTrack
|
AnboxStream::_onExtraVideoTrack
|
AnboxStream::_onExtraVideoTrack
|
AnboxStream::._options.callbacks.videoTrackAdded(trackIndex);
|
So the consumer(e.g. dashboard) could call AnboxStrae::attachDisplay to attach the display to a video container once the signal is fired.

I just polished that a bit.
Please take another look.

js/anbox-stream-sdk.js
adglkh: It's a private API exclusively used in _onPointerEvent, where displayId defaults to 0 anyway.
So it should be fine.

js/anbox-stream-sdk.js
adglkh: See above comment.

tarek-hanna (approved):
Thanks Gary 👍  LGTM

js/anbox-stream-sdk.js
lorumic: Yes, it was that "External displays" that felt out of place and left me puzzled. Without it, it's clear. :+1:

js/anbox-stream-sdk.js
lorumic: @adglkh We still have a fixed black background set at line 1212.

lorumic (changes requested):
Thanks for addressing the comments and fixing the open issues! There is only one leftover black bg colour to be addressed (see [open comment above](https://github.com/canonical/anbox-streaming-sdk/pull/219#discussion_r3611561006)), then we should be good to merge 👍

adglkh:
> Thanks for addressing the comments and fixing the open issues! There is only one leftover black bg colour to be addressed (see [open comment above](https://github.com/canonical/anbox-streaming-sdk/pull/219#discussion_r3611561006)), then we should be good to merge 👍

A: oh, I missed that out. I just fixed that.
    Thanks for pointing this out.

js/anbox-stream-sdk.js
adglkh: I missed that out and just fixed that.
Please take another look

lorumic (approved):
Perfect, LGTM 👍

PR #224, opened 2026-07-24
Support upscaling feature for multi displays

adglkh:
## Done

This change applies WebGL based upscaling filter to all displays once
experimental.upscaling option is enabled when a session streams with
multi displays. In such a case, this display is rendered through a
canvas instead of showing the <video> element directly, and the video
element stays in the DOM (hidden) as the source texture.

## QA

1. Import the JS SDK into a stream client
2. Register the statsUpdate callback function to the AnboxStream
```
        stream = new AnboxStream({
          connector: connector,
          ...
          experimental: {
              upscaling: {
                enabled:  true,
              },
          ...
       }
```

4. Launch an instance with multi displays form euphotic image:
   $ amc launch resolute:android16-cf:amd64 --enable-streaming --devmode -c 4 -m 8GB --name test-multi-display \
    --display=width=1280,height=720,fps=60 \
    --display=width=480,height=640,dpi=120 \
5. Once the stream is up and running,  check if the upscaling filter is applied for all three displays
 
## JIRA / Launchpad bug

Fixes https://warthogs.atlassian.net/browse/AC-4913

## Documentation

A: yes, this feature will be documented as part of multi display feature introduced in 1.31.0 release.

### DIFF
--- a/js/tests/unit/multi-display.test.js
+++ b/js/tests/unit/multi-display.test.js
@@ -77,6 +77,9 @@ beforeEach(() => {
   window.HTMLCanvasElement.prototype.getContext = () => {
     return {};
   };
+  // jsdom does not implement media playback. Calling play() on hidden video
+  // elements logs "Not implemented: HTMLMediaElement.prototype.play" error.
+  window.HTMLMediaElement.prototype.play = () => Promise.resolve();
   addContainer("main-container");
   addContainer("cell-1");
   addContainer("cell-2");
--- a/js/tests/unit/multi-display.test.js
+++ b/js/tests/unit/multi-display.test.js
@@ -294,8 +294,7 @@ describe("attachDisplay (dynamic mode)", () => {
 
     stream.attachDisplay(1, "cell-1");
 
-    const videoId = `${stream._videoID}-display-1`;
-    const video = document.getElementById(videoId);
+    const video = document.getElementById(stream._videoIdFor(1));
     expect(video).not.toBeNull();
     expect(video.srcObject).toBe(fakeStream);
     expect(stream._pendingVideoTracks[1]).toBeUndefined();
@@ -372,15 +371,18 @@ describe("attachDisplay with upscaling enabled", () => {
     });
   }
 
-  test("creates a canvas for an additional display and hides its video element", () => {
+  function setupStreamCanvasAndAttachDisplay() {
     const stream = makeUpscalingStream();
     stubStreamCanvasCreation(stream);
     stream._pendingVideoTracks[1] = makeFakeStream();
-
     stream.attachDisplay(1, "cell-1");
-
-    const video = document.getElementById(`${stream._videoID}-display-1`);
+    const video = document.getElementById(stream._videoIdFor(1));
     const canvas = document.getElementById(stream._canvasIdFor(1));
+    return { stream, video, canvas };
+  }
+
+  test("creates a canvas for an additional display and hides its video element", () => {
+    const { stream, video, canvas } = setupStreamCanvasAndAttachDisplay();
     expect(video.style.display).toBe("none");
     expect(canvas).not.toBeNull();
     expect(canvas.parentElement).toBe(document.getElementById("cell-1"));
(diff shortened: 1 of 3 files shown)


js/tests/unit/multi-display.test.js
```diff
@@ -341,6 +349,105 @@ describe("attachDisplay (dynamic mode)", () => {
...
+  }
+
+  test("creates a canvas for an additional display and hides its video element", () => {
+    const stream = makeUpscalingStream();
+    stubStreamCanvasCreation(stream);
+    stream._pendingVideoTracks[1] = makeFakeStream();
+
+    stream.attachDisplay(1, "cell-1");
+
+    const video = document.getElementById(`${stream._videoID}-display-1`);
```
lorumic: We could use `_videoIdFor` here:

```suggestion
    const video = document.getElementById(`${stream._videoIdFor(1)`);
```

js/tests/unit/multi-display.test.js
```diff
@@ -341,6 +349,105 @@ describe("attachDisplay (dynamic mode)", () => {
...
+    expect(stream._streamCanvases[1]).toBeDefined();
+  });
+
+  test("starts canvas rendering once the video's metadata is loaded", () => {
+    const stream = makeUpscalingStream();
+    stubStreamCanvasCreation(stream);
+    stream._pendingVideoTracks[1] = makeFakeStream();
+
+    stream.attachDisplay(1, "cell-1");
+    const video = document.getElementById(`${stream._videoID}-display-1`);
```
lorumic: And here as well:

```suggestion
    const video = document.getElementById(`${stream._videoIdFor(1)`);
```

js/tests/unit/multi-display.test.js
```diff
@@ -341,6 +349,105 @@ describe("attachDisplay (dynamic mode)", () => {
...
+    expect(stream._streamCanvases[1].startRendering).toHaveBeenCalledTimes(1);
+  });
+
+  test("_computeMultiDisplayDimensions sizes the canvas, not the hidden video", () => {
+    const stream = makeUpscalingStream();
+    stubStreamCanvasCreation(stream);
+    stream._pendingVideoTracks[1] = makeFakeStream();
+    stream.attachDisplay(1, "cell-1");
+
+    const video = document.getElementById(`${stream._videoID}-display-1`);
```
lorumic:
```suggestion
    const video = document.getElementById(`${stream._videoIdFor(1)`);
```

js/tests/unit/multi-display.test.js
```diff
@@ -341,6 +349,105 @@ describe("attachDisplay (dynamic mode)", () => {
...
+
+    stream._computeMultiDisplayDimensions(
+      video,
+      document.getElementById("cell-1"),
+      1,
+      canvas,
+    );
+
+    expect(canvas.style.width).not.toBe("");
+    expect(canvas.style.height).not.toBe("");
```
lorumic: Can we convert this into a positive expectation, and expect exact values instead?

```suggestion
    expect(canvas.style.width).toBe(400);
    expect(canvas.style.height).toBe(800);
```

js/tests/unit/multi-display.test.js
lorumic: These 5 lines are repeated unchanged in 3 out of 4 tests currently. Shall we create a reusable function that returns `{ stream, video }` so that we can avoid the code duplication?

lorumic (approved):
Looks very good overall, thanks again for fixing the autoplay and rendering issues so quickly.
I'm leaving a few optional comments on the test code - mostly nitpicks, feel free to ignore them and proceed with the merge if you want.

js/tests/unit/multi-display.test.js
adglkh: Sure, updated this.

js/tests/unit/multi-display.test.js
adglkh: +1, thanks for pointing this out.

js/tests/unit/multi-display.test.js
adglkh: Make sense to me.
Fixed that.
Please take another look

ajanon (approved):
LGTM

PR #242, opened 2026-09-04
Stop client from creating its own control channel

adglkh:
## Done

For API version >= 2 the server acts as the SDP offerer and already creates the "control" data channel in configure_peer_connection() before generating its offer. The client SDK was independently calling createDataChannel("control") too, resulting in two distinct control channels that only happened to share a label. The server's PeerConnection::OnDataChannel() matches channels by label and would overwrite its tracked channel and re-register the existing observer onto the new channel, causing resume_from_last_state() (and other control-channel-open handling) to fire twice per connection.

Hence stop client from creating its own control channel when API version >=2.

NOTE:
1. unlike the control channel, oob data channels are always created locally by the client. The server relies on seeing the client's channel arrive to know when to start forwarding data for it. So this is intentionally left unchanged.
2. this is a client-only change and does not requires server-side changes and works against any server version.

## QA

All tests should be pass.
This change should be fully backward & forward compatibility.
Meaning nothing would break for all old images prior to 1.31, or new images.

## JIRA / Launchpad bug

Fixes #

## Documentation

### DIFF
--- a/js/anbox-stream-sdk.js
+++ b/js/anbox-stream-sdk.js
@@ -3586,6 +3586,7 @@ class AnboxWebRTCManager {
     this._pc.oniceconnectionstatechange =
       this._onRtcIceConnectionStateChange.bind(this);
     this._pc.onicecandidate = this._onRtcIceCandidate.bind(this);
+    this._pc.ondatachannel = this._onRtcDataChannel.bind(this);
 
     this._createControlChannel();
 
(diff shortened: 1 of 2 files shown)


lorumic (approved):
Tested the changes in the dashboard with both new and old images: no regression was found. LGTM 👍

adglkh:
> Tested the changes in the dashboard with both new and old images: no regression was found. LGTM 👍

Thanks!

PR #243, opened 2026-09-17
Fire onReady only once and wait for all displays

adglkh:
## Done

AnboxWebRTCManager._onRtcTrack() evaluated its readiness condition on
every incoming track (extra display video tracks, or audio arriving
after video), which could call _onReady() multiple times per session.
The correct behavior is to trigger the readiness condition only after
every expected display's track has arrived, preventing redundant
callbacks that could break caller logic.

## QA

1 . Launch an instance with multi displays form euphotic image:

    $ amc launch resolute:android16-cf:amd64 --enable-streaming \
        --devmode -c 4 -m 8GB --name test-multi-display \
        --display=width=1280,height=720,fps=60 \
        --display=width=480,height=640,dpi=120 \

2. Join the session from the browser and ensure the callback function `onReady` 
    from the AnboxStream options  is fired only once

## JIRA / Launchpad bug

Fixes #

## Documentation

A: no, it doesn't


lorumic (approved):
Tested in the dashboard, I confirm that `onReady` gets now called once instead of N times (where N is the number of displays). Approved 👍

adglkh:
> Tested in the dashboard, I confirm that `onReady` gets now called once instead of N times (where N is the number of displays). Approved 👍

A: thanks for the testing!

